PR #243, opened 2025-07-28
fixed accessability of disabled tenant

HanifAbRazak:


XiansAi.Server.Src/Features/WebApi/Services/TenantService.cs
```diff
@@ -220,6 +220,12 @@ public async Task<ServiceResult<Tenant>> GetCurrentTenantInfo()
                 return ServiceResult<Tenant>.BadRequest("Tenant ID is not set in the context");
             }
 
+            // handle default tenant
```
hasith, line 223: Why do we send null result in case of the "default" tenant? In the default tenant I was experiencing issues due to this. Any particular reason?

PR #385, opened 2026-03-16
Secret Vault Key + Scope Only & Tenant Existence Validation and also Ensuring scope level key uniqueness

kavishdiv99nanayakkara:
## Summary

Secret Vault APIs now identify secrets by **key + scope** only for fetch, update, and delete. The “get by id” endpoint has been removed. When a tenant scope is used, the server validates that the tenant exists in the Xians tenant collection.

### Service (`SecretVaultService`)

- **Added** `UpdateByKeyAsync(string key, SecretVaultUpdateInput input, string actorUserId)` – finds secret via `FindForAccessAsync(key, scope)` then delegates to `UpdateAsync(entity.Id, ...)`.
- **Added** `DeleteByKeyAsync(string key, string? tenantId, string? agentId, string? userId, string? activationName)` – finds via `FindForAccessAsync` then calls `DeleteAsync(entity.Id)`.
- Existing `GetByIdAsync`, `UpdateAsync`, `DeleteAsync` remain for internal use.

### Agent API (`SecretVaultEndpoints`)

- **Removed** `GET /api/agent/secrets/{id}` (Get by id).
- **Update**: `PUT /api/agent/secrets` – body must include `key`; scope from body, resolved with `TryResolveScope`; tenant existence checked when effective tenant is set; calls `UpdateByKeyAsync(request.Key, ...)`.
- **Delete**: `DELETE /api/agent/secrets?key=...&tenantId=...&agentId=...&userId=...&activationName=...` – scope from query, resolved with `TryResolveScope`; calls `DeleteByKeyAsync(key, effectiveTenantId, ...)`.

### Admin API (`AdminSecretVaultEndpoints`)

- **Removed** `GET /api/v1/admin/secrets/{id}` (Get by id).
- **Update**: `PUT /api/v1/admin/secrets` – body must include `key`; scope and tenant validation (when `request.TenantId` set) then `UpdateByKeyAsync(request.Key, ...)`.
- **Delete**: `DELETE /api/v1/admin/secrets?key=...&tenantId=...&agentId=...&userId=...&activationName=...` – scope from query, `TryResolveScope`, then `DeleteByKeyAsync`.

### Tenant existence validation

- **Agent API**: On create, update, and fetch, when effective `tenantId` is non-empty, endpoint calls `ITenantCacheService.GetByTenantIdAsync`; if tenant is missing, returns **404** “Tenant not found”.
- **Admin API**: On create and update, when `request.TenantId` and effective tenant are set, same check; 404 if tenant does not exist.

### Documentation

- **SECRET_VAULT.md**: Data flow (no “Get by id”), API reference table (PUT/DELETE by key+scope, no GET by id), create response note that subsequent operations use key+scope.
- **SECRET_VAULT_VALIDATION.md**: `CanAccessSecretTenant` described as internal; Agent/Admin tables updated for update/delete by key+scope.

### HTTP test files

- **agent-secret-vault-endpoints.http**: Removed `@secretId` and id-based GET/PUT/DELETE; added PUT body with `key`, DELETE with `?key=...`.
- **admin-secret-vault-endpoints.http**: Same; PUT with `key` in body, DELETE with key and scope query params.

## Breaking changes

- Clients must not rely on `GET /api/v1/admin/secrets/{id}` or `GET /api/agent/secrets/{id}`; use list + fetch by key (or list only) instead.
- Update and delete must be called with key and correct scope (query or body as above).

## Checklist

- [ ] DB and uniqueness unchanged (key + scope tuple).
- [ ] Tenant validation only when tenant scope is present; cross-tenant (null) unchanged.

### DIFF
--- a/XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs
+++ b/XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs
@@ -118,34 +118,15 @@ public static class AdminSecretVaultEndpoints
         .WithName("FetchSecretByKey")
         .WithOpenApi(o => { o.Summary = "Fetch secret by key"; o.Description = "Returns decrypted value and optional AdditionalData. Scope by tenantId, agentId, userId, activationName."; return o; });
 
-        group.MapGet("/{id}", async (
-            string id,
-            [FromServices] ISecretVaultService service,
-            [FromServices] ITenantContext tenantContext) =>
-        {
-            var result = await service.GetByIdAsync(id);
-            if (!result.IsSuccess || result.Data == null)
-                return result.ToHttpResult();
-            if (!SecretVaultScopeEnforcement.CanAccessSecretTenant(tenantContext, result.Data.TenantId))
-                return ServiceResult<SecretVaultGetResponse?>.Forbidden("Access denied. Secret is not in your tenant.").ToHttpResult();
-            return result.ToHttpResult();
-        })
-        .WithName("GetSecretById")
-        .WithOpenApi(o => { o.Summary = "Get secret by id"; return o; });
-
-        group.MapPut("/{id}", async (
-            string id,
+        group.MapPut("", async (
             [FromBody] SecretVaultUpdateRequest request,
             [FromServices] ISecretVaultService service,
             [FromServices] ITenantContext tenantContext,
             [FromServices] ITenantCacheService tenantCacheService,
             CancellationToken cancellationToken) =>
         {
-            var getResult = await service.GetByIdAsync(id);
-            if (!getResult.IsSuccess || getResult.Data == null)
-                return getResult.ToHttpResult();
-            if (!SecretVaultScopeEnforcement.CanAccessSecretTenant(tenantContext, getResult.Data.TenantId))
-                return ServiceResult<SecretVaultGetResponse>.Forbidden("Access denied. Secret is not in your tenant.").ToHttpResult();
+            if (string.IsNullOrWhiteSpace(request.Key))
+                return ServiceResult<SecretVaultGetResponse>.BadRequest("Key is required").ToHttpResult();
 
             if (!SecretVaultScopeEnforcement.TryResolveScope(
                 tenantContext,
@@ -180,27 +161,44 @@ public static class AdminSecretVaultEndpoints
                 effectiveUserId,
                 effectiveActivationName,
                 SecretVaultService.NormalizeAdditionalDataFromRequest(request.AdditionalData));
-            var result = await service.UpdateAsync(id, input, actor);
+            var result = await service.UpdateByKeyAsync(request.Key, input, actor);
             return result.ToHttpResult();
         })
-        .WithName("UpdateSecret")
-        .WithOpenApi(o => { o.Summary = "Update secret"; return o; });
+        .WithName("UpdateSecretByKey")
+        .WithOpenApi(o => { o.Summary = "Update secret by key and scope"; return o; });
 
-        group.MapDelete("/{id}", async (
-            string id,
+        group.MapDelete("", async (
+            [FromQuery] string key,
+            [FromQuery] string? tenantId,
+            [FromQuery] string? agentId,
+            [FromQuery] string? userId,
+            [FromQuery] string? activationName,
             [FromServices] ISecretVaultService service,
             [FromServices] ITenantContext tenantContext) =>
         {
-            var getResult = await service.GetByIdAsync(id);
-            if (!getResult.IsSuccess)
-                return getResult.ToHttpResult();
-            if (getResult.Data != null && !SecretVaultScopeEnforcement.CanAccessSecretTenant(tenantContext, getResult.Data.TenantId))
-                return ServiceResult<bool>.Forbidden("Access denied. Secret is not in your tenant.").ToHttpResult();
-            var result = await service.DeleteAsync(id);
+            if (string.IsNullOrWhiteSpace(key))
+                return ServiceResult<bool>.BadRequest("Key is required").ToHttpResult();
+
+            if (!SecretVaultScopeEnforcement.TryResolveScope(
+                tenantContext,
+                tenantId,
+                agentId,
+                userId,
+                activationName,
+                out var effectiveTenantId,
+                out var effectiveAgentId,
+                out var effectiveUserId,
+                out var effectiveActivationName,
+                out var forbiddenResult))
+            {
+                return forbiddenResult!.ToHttpResult();
+            }
+
+            var result = await service.DeleteByKeyAsync(key, effectiveTenantId, effectiveAgentId, effectiveUserId, effectiveActivationName);
             return result.ToHttpResult();
         })
-        .WithName("DeleteSecret")
-        .WithOpenApi(o => { o.Summary = "Delete secret"; return o; });
+        .WithName("DeleteSecretByKey")
+        .WithOpenApi(o => { o.Summary = "Delete secret by key and scope"; return o; });
     }
 }
 
--- a/XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs
+++ b/XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs
@@ -217,6 +217,7 @@ public class SecretVaultCreateRequest
 
 public class SecretVaultUpdateRequest
 {
+    public string Key { get; set; } = null!;
     public string? Value { get; set; }
     public string? TenantId { get; set; }
     public string? AgentId { get; set; }
--- a/XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs
+++ b/XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs
@@ -16,7 +16,9 @@ public static class AdminSecretVaultEndpoints
         group.MapPost("", async (
             [FromBody] SecretVaultCreateRequest request,
             [FromServices] ISecretVaultService service,
-            [FromServices] ITenantContext tenantContext) =>
+            [FromServices] ITenantContext tenantContext,
+            [FromServices] ITenantCacheService tenantCacheService,
+            CancellationToken cancellationToken) =>
         {
             if (!SecretVaultScopeEnforcement.TryResolveScope(
                 tenantContext,
@@ -31,6 +33,16 @@ public static class AdminSecretVaultEndpoints
                 out var forbiddenResult))
             {
                 return forbiddenResult!.ToHttpResult();
+            }
+
+            // If a tenant scope is provided/resolved, ensure the tenant exists when tenantId scope is explicitly set.
+            if (!string.IsNullOrWhiteSpace(request.TenantId) && !string.IsNullOrWhiteSpace(effectiveTenantId))
+            {
+                var tenant = await tenantCacheService.GetByTenantIdAsync(effectiveTenantId, cancellationToken);
+                if (tenant == null)
+                {
+                    return ServiceResult<SecretVaultGetResponse>.NotFound("Tenant not found").ToHttpResult();
+                }
             }
 
             var actor = tenantContext.LoggedInUser ?? "system";
@@ -125,7 +137,9 @@ public static class AdminSecretVaultEndpoints
             string id,
             [FromBody] SecretVaultUpdateRequest request,
             [FromServices] ISecretVaultService service,
-            [FromServices] ITenantContext tenantContext) =>
+            [FromServices] ITenantContext tenantContext,
+            [FromServices] ITenantCacheService tenantCacheService,
+            CancellationToken cancellationToken) =>
         {
             var getResult = await service.GetByIdAsync(id);
             if (!getResult.IsSuccess || getResult.Data == null)
@@ -146,6 +160,16 @@ public static class AdminSecretVaultEndpoints
                 out var forbiddenResult))
             {
                 return forbiddenResult!.ToHttpResult();
+            }
+
+            // If a tenant scope is provided/resolved, ensure the tenant exists when tenantId scope is explicitly set.
+            if (!string.IsNullOrWhiteSpace(request.TenantId) && !string.IsNullOrWhiteSpace(effectiveTenantId))
+            {
+                var tenant = await tenantCacheService.GetByTenantIdAsync(effectiveTenantId, cancellationToken);
+                if (tenant == null)
+                {
+                    return ServiceResult<SecretVaultGetResponse>.NotFound("Tenant not found").ToHttpResult();
+                }
             }
 
             var actor = tenantContext.LoggedInUser ?? "system";
--- a/XiansAi.Server.Src/Features/AgentApi/Endpoints/SecretVaultEndpoints.cs
+++ b/XiansAi.Server.Src/Features/AgentApi/Endpoints/SecretVaultEndpoints.cs
@@ -230,6 +230,7 @@ public class AgentSecretVaultCreateRequest
 
 public class AgentSecretVaultUpdateRequest
 {
+    public string Key { get; set; } = null!;
     public string? Value { get; set; }
     public string? TenantId { get; set; }
     public string? AgentId { get; set; }
--- a/XiansAi.Server.Src/Features/AgentApi/Endpoints/SecretVaultEndpoints.cs
+++ b/XiansAi.Server.Src/Features/AgentApi/Endpoints/SecretVaultEndpoints.cs
@@ -131,34 +131,15 @@ public static class SecretVaultEndpoints
         .WithName("Agent_FetchSecretByKey")
         .WithOpenApi(o => { o.Summary = "Fetch secret by key"; o.Description = "Returns decrypted value and optional AdditionalData. Scope by tenantId, agentId, userId, activationName."; return o; });
 
-        group.MapGet("/{id}", async (
-            string id,
-            [FromServices] ISecretVaultService service,
-            [FromServices] ITenantContext tenantContext) =>
-        {
-            var result = await service.GetByIdAsync(id);
-            if (!result.IsSuccess || result.Data == null)
-                return result.ToHttpResult();
-            if (!SecretVaultScopeEnforcement.CanAccessSecretTenant(tenantContext, result.Data.TenantId))
-                return ServiceResult<SecretVaultGetResponse?>.Forbidden("Access denied. Secret is not in your tenant.").ToHttpResult();
-            return result.ToHttpResult();
-        })
-        .WithName("Agent_GetSecretById")
-        .WithOpenApi(o => { o.Summary = "Get secret by id"; return o; });
-
-        group.MapPut("/{id}", async (
-            string id,
+        group.MapPut("", async (
             [FromBody] AgentSecretVaultUpdateRequest request,
             [FromServices] ISecretVaultService service,
             [FromServices] ITenantContext tenantContext,
             [FromServices] ITenantCacheService tenantCacheService,
             CancellationToken cancellationToken) =>
         {
-            var getResult = await service.GetByIdAsync(id);
-            if (!getResult.IsSuccess || getResult.Data == null)
-                return getResult.ToHttpResult();
-            if (!SecretVaultScopeEnforcement.CanAccessSecretTenant(tenantContext, getResult.Data.TenantId))
-                return ServiceResult<SecretVaultGetResponse>.Forbidden("Access denied. Secret is not in your tenant.").ToHttpResult();
+            if (string.IsNullOrWhiteSpace(request.Key))
+                return ServiceResult<SecretVaultGetResponse>.BadRequest("Key is required").ToHttpResult();
 
             if (!SecretVaultScopeEnforcement.TryResolveScope(
                 tenantContext,
@@ -193,27 +174,44 @@ public static class SecretVaultEndpoints
                 effectiveUserId,
                 effectiveActivationName,
                 SecretVaultService.NormalizeAdditionalDataFromRequest(request.AdditionalData));
-            var result = await service.UpdateAsync(id, input, actor);
+            var result = await service.UpdateByKeyAsync(request.Key, input, actor);
             return result.ToHttpResult();
         })
-        .WithName("Agent_UpdateSecret")
-        .WithOpenApi(o => { o.Summary = "Update secret"; return o; });
+        .WithName("Agent_UpdateSecretByKey")
+        .WithOpenApi(o => { o.Summary = "Update secret by key and scope"; return o; });
 
-        group.MapDelete("/{id}", async (
-            string id,
+        group.MapDelete("", async (
+            [FromQuery] string key,
+            [FromQuery] string? tenantId,
+            [FromQuery] string? agentId,
+            [FromQuery] string? userId,
+            [FromQuery] string? activationName,
             [FromServices] ISecretVaultService service,
             [FromServices] ITenantContext tenantContext) =>
         {
-            var getResult = await service.GetByIdAsync(id);
-            if (!getResult.IsSuccess)
-                return getResult.ToHttpResult();
-            if (getResult.Data != null && !SecretVaultScopeEnforcement.CanAccessSecretTenant(tenantContext, getResult.Data.TenantId))
-                return ServiceResult<bool>.Forbidden("Access denied. Secret is not in your tenant.").ToHttpResult();
-            var result = await service.DeleteAsync(id);
+            if (string.IsNullOrWhiteSpace(key))
+                return ServiceResult<bool>.BadRequest("Key is required").ToHttpResult();
+
+            if (!SecretVaultScopeEnforcement.TryResolveScope(
+                tenantContext,
+                tenantId,
+                agentId,
+                userId,
+                activationName,
+                out var effectiveTenantId,
+                out var effectiveAgentId,
+                out var effectiveUserId,
+                out var effectiveActivationName,
+                out var forbiddenResult))
+            {
+                return forbiddenResult!.ToHttpResult();
+            }
+
+            var result = await service.DeleteByKeyAsync(key, effectiveTenantId, effectiveAgentId, effectiveUserId, effectiveActivationName);
             return result.ToHttpResult();
         })
-        .WithName("Agent_DeleteSecret")
-        .WithOpenApi(o => { o.Summary = "Delete secret"; return o; });
+        .WithName("Agent_DeleteSecretByKey")
+        .WithOpenApi(o => { o.Summary = "Delete secret by key and scope"; return o; });
     }
 }
 
--- a/XiansAi.Server.Src/Features/AgentApi/Endpoints/SecretVaultEndpoints.cs
+++ b/XiansAi.Server.Src/Features/AgentApi/Endpoints/SecretVaultEndpoints.cs
@@ -17,7 +17,9 @@ public static class SecretVaultEndpoints
         group.MapPost("", async (
             [FromBody] AgentSecretVaultCreateRequest request,
             [FromServices] ISecretVaultService service,
-            [FromServices] ITenantContext tenantContext) =>
+            [FromServices] ITenantContext tenantContext,
+            [FromServices] ITenantCacheService tenantCacheService,
+            CancellationToken cancellationToken) =>
         {
             if (!SecretVaultScopeEnforcement.TryResolveScope(
                 tenantContext,
@@ -32,6 +34,16 @@ public static class SecretVaultEndpoints
                 out var forbiddenResult))
             {
                 return forbiddenResult!.ToHttpResult();
+            }
+
+            // If a tenant scope is provided/resolved, ensure the tenant exists.
+            if (!string.IsNullOrWhiteSpace(effectiveTenantId))
+            {
+                var tenant = await tenantCacheService.GetByTenantIdAsync(effectiveTenantId, cancellationToken);
+                if (tenant == null)
+                {
+                    return ServiceResult<SecretVaultGetResponse>.NotFound("Tenant not found").ToHttpResult();
+                }
             }
 
             var actor = tenantContext.LoggedInUser ?? "system";
@@ -84,7 +96,9 @@ public static class SecretVaultEndpoints
             [FromQuery] string? userId,
             [FromQuery] string? activationName,
             [FromServices] ISecretVaultService service,
-            [FromServices] ITenantContext tenantContext) =>
+            [FromServices] ITenantContext tenantContext,
+            [FromServices] ITenantCacheService tenantCacheService,
+            CancellationToken cancellationToken) =>
         {
             if (!SecretVaultScopeEnforcement.TryResolveScope(
                 tenantContext,
@@ -99,6 +113,16 @@ public static class SecretVaultEndpoints
                 out var forbiddenResult))
             {
                 return forbiddenResult!.ToHttpResult();
+            }
+
+            // If a tenant scope is provided/resolved, ensure the tenant exists.
+            if (!string.IsNullOrWhiteSpace(effectiveTenantId))
+            {
+                var tenant = await tenantCacheService.GetByTenantIdAsync(effectiveTenantId, cancellationToken);
+                if (tenant == null)
+                {
+                    return ServiceResult<SecretVaultFetchResponse?>.NotFound("Tenant not found").ToHttpResult();
+                }
             }
 
             var result = await service.FetchByKeyAsync(key, effectiveTenantId, effectiveAgentId, effectiveUserId, effectiveActivationName);
@@ -126,7 +150,9 @@ public static class SecretVaultEndpoints
             string id,
             [FromBody] AgentSecretVaultUpdateRequest request,
             [FromServices] ISecretVaultService service,
-            [FromServices] ITenantContext tenantContext) =>
+            [FromServices] ITenantContext tenantContext,
+            [FromServices] ITenantCacheService tenantCacheService,
+            CancellationToken cancellationToken) =>
         {
             var getResult = await service.GetByIdAsync(id);
             if (!getResult.IsSuccess || getResult.Data == null)
@@ -147,6 +173,16 @@ public static class SecretVaultEndpoints
                 out var forbiddenResult))
             {
                 return forbiddenResult!.ToHttpResult();
+            }
+
+            // If a tenant scope is provided/resolved, ensure the tenant exists.
+            if (!string.IsNullOrWhiteSpace(effectiveTenantId))
+            {
+                var tenant = await tenantCacheService.GetByTenantIdAsync(effectiveTenantId, cancellationToken);
+                if (tenant == null)
+                {
+                    return ServiceResult<SecretVaultGetResponse>.NotFound("Tenant not found").ToHttpResult();
+                }
             }
 
             var actor = tenantContext.LoggedInUser ?? "system";
--- a/XiansAi.Server.Src/Shared/Repositories/SecretVaultRepository.cs
+++ b/XiansAi.Server.Src/Shared/Repositories/SecretVaultRepository.cs
@@ -9,6 +9,7 @@ public interface ISecretVaultRepository
     Task<SecretVault?> GetByIdAsync(string id);
     Task<SecretVault?> GetByKeyAsync(string key);
     Task<bool> ExistsByKeyAsync(string key);
+    Task<bool> ExistsByKeyAndScopeAsync(string key, string? tenantId, string? agentId, string? userId, string? activationName, string? excludeId = null);
     Task<SecretVault?> FindForAccessAsync(string key, string? tenantId, string? agentId, string? userId, string? activationName);
     Task<List<SecretVault>> ListAsync(string? tenantId, string? agentId, string? activationName);
     Task CreateAsync(SecretVault entity);
@@ -76,32 +77,7 @@ public class SecretVaultRepository : ISecretVaultRepository
     {
         try
         {
-            var keyFilter = Builders<SecretVault>.Filter.Eq(x => x.Key, key);
-            var filters = new List<MongoDB.Driver.FilterDefinition<SecretVault>> { keyFilter };
-
-            // Request scope must match document scope: if request omits scope, document must have that scope null.
-            if (!string.IsNullOrWhiteSpace(tenantId))
-                filters.Add(Builders<SecretVault>.Filter.Eq(x => x.TenantId, tenantId));
-            else
-                filters.Add(Builders<SecretVault>.Filter.Eq(x => x.TenantId, (string?)null));
-
-            if (!string.IsNullOrWhiteSpace(agentId))
-                filters.Add(Builders<SecretVault>.Filter.Eq(x => x.AgentId, agentId));
-            else
-                filters.Add(Builders<SecretVault>.Filter.Eq(x => x.AgentId, (string?)null));
-
-            if (!string.IsNullOrWhiteSpace(userId))
-                filters.Add(Builders<SecretVault>.Filter.Eq(x => x.UserId, userId));
-            else
-                filters.Add(Builders<SecretVault>.Filter.Eq(x => x.UserId, (string?)null));
-
-            // Request scope must match document scope: if request omits scope, document must have that scope null.
-            if (!string.IsNullOrWhiteSpace(activationName))
-                filters.Add(Builders<SecretVault>.Filter.Eq(x => x.ActivationName, activationName));
-            else
-                filters.Add(Builders<SecretVault>.Filter.Eq(x => x.ActivationName, (string?)null));
-
-            var filter = Builders<SecretVault>.Filter.And(filters);
+            var filter = BuildKeyAndScopeFilter(key, tenantId, agentId, userId, activationName);
             return await _collection.Find(filter).FirstOrDefaultAsync();
         }
         catch (Exception ex)
@@ -109,6 +85,55 @@ public class SecretVaultRepository : ISecretVaultRepository
             _logger.LogError(ex, "Error finding secret for access key {Key}", key);
             return null;
         }
+    }
+
+    public async Task<bool> ExistsByKeyAndScopeAsync(string key, string? tenantId, string? agentId, string? userId, string? activationName, string? excludeId = null)
+    {
+        try
+        {
+            var filter = BuildKeyAndScopeFilter(key, tenantId, agentId, userId, activationName);
+            if (!string.IsNullOrEmpty(excludeId))
+            {
+                var builder = Builders<SecretVault>.Filter;
+                filter = builder.And(filter, builder.Ne(x => x.Id, excludeId));
+            }
+
+            var count = await _collection.CountDocumentsAsync(filter);
+            return count > 0;
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Error checking existence for key {Key} and scope", key);
+            return false;
+        }
+    }
+
+    private static FilterDefinition<SecretVault> BuildKeyAndScopeFilter(string key, string? tenantId, string? agentId, string? userId, string? activationName)
+    {
+        var builder = Builders<SecretVault>.Filter;
+        var filters = new List<FilterDefinition<SecretVault>>
+        {
+            builder.Eq(x => x.Key, key)
+        };
+
+        // Request scope must match document scope: if request omits scope, document must have that scope null.
+        filters.Add(!string.IsNullOrWhiteSpace(tenantId)
+            ? builder.Eq(x => x.TenantId, tenantId)
+            : builder.Eq(x => x.TenantId, (string?)null));
+
+        filters.Add(!string.IsNullOrWhiteSpace(agentId)
+            ? builder.Eq(x => x.AgentId, agentId)
+            : builder.Eq(x => x.AgentId, (string?)null));
+
+        filters.Add(!string.IsNullOrWhiteSpace(userId)
+            ? builder.Eq(x => x.UserId, userId)
+            : builder.Eq(x => x.UserId, (string?)null));
+
+        filters.Add(!string.IsNullOrWhiteSpace(activationName)
+            ? builder.Eq(x => x.ActivationName, activationName)
+            : builder.Eq(x => x.ActivationName, (string?)null));
+
+        return builder.And(filters);
     }
 
     public async Task<List<SecretVault>> ListAsync(string? tenantId, string? agentId, string? activationName)
--- a/XiansAi.Server.Src/Shared/Services/SecretVaultService.cs
+++ b/XiansAi.Server.Src/Shared/Services/SecretVaultService.cs
@@ -104,9 +104,20 @@ public class SecretVaultService : ISecretVaultService
         if (string.IsNullOrWhiteSpace(input.Value))
             return ServiceResult<SecretVaultGetResponse>.BadRequest("Value is required");
 
-        var exists = await _repository.ExistsByKeyAsync(input.Key);
+        // Normalize scope values so that uniqueness is enforced on the composite (Key + scope) with consistent null handling.
+        var normalizedTenantId = string.IsNullOrWhiteSpace(input.TenantId) ? null : input.TenantId;
+        var normalizedAgentId = string.IsNullOrWhiteSpace(input.AgentId) ? null : input.AgentId;
+        var normalizedUserId = string.IsNullOrWhiteSpace(input.UserId) ? null : input.UserId;
+        var normalizedActivationName = string.IsNullOrWhiteSpace(input.ActivationName) ? null : input.ActivationName;
+
+        var exists = await _repository.ExistsByKeyAndScopeAsync(
+            input.Key,
+            normalizedTenantId,
+            normalizedAgentId,
+            normalizedUserId,
+            normalizedActivationName);
         if (exists)
-            return ServiceResult<SecretVaultGetResponse>.Conflict("A secret with this key already exists");
+            return ServiceResult<SecretVaultGetResponse>.Conflict("A secret with this key already exists for the same scope");
 
         var (sanitizedAdditionalData, additionalDataError) = ValidateAndSanitizeAdditionalData(input.AdditionalData);
         if (additionalDataError != null)
@@ -121,10 +132,10 @@ public class SecretVaultService : ISecretVaultService
                 Id = MongoDB.Bson.ObjectId.GenerateNewId().ToString(),
                 Key = input.Key,
                 EncryptedValue = encrypted,
-                TenantId = string.IsNullOrWhiteSpace(input.TenantId) ? null : input.TenantId,
-                AgentId = string.IsNullOrWhiteSpace(input.AgentId) ? null : input.AgentId,
-                UserId = string.IsNullOrWhiteSpace(input.UserId) ? null : input.UserId,
-                ActivationName = string.IsNullOrWhiteSpace(input.ActivationName) ? null : input.ActivationName,
+                TenantId = normalizedTenantId,
+                AgentId = normalizedAgentId,
+                UserId = normalizedUserId,
+                ActivationName = normalizedActivationName,
                 AdditionalData = sanitizedAdditionalData,
                 CreatedAt = now,
                 CreatedBy = actorUserId
@@ -215,6 +226,17 @@ public class SecretVaultService : ISecretVaultService
                     return ServiceResult<SecretVaultGetResponse>.BadRequest(additionalDataError);
                 entity.AdditionalData = sanitizedAdditionalData;
             }
+
+            // Enforce uniqueness on (Key + scope) combination, excluding the current document.
+            var existsForScope = await _repository.ExistsByKeyAndScopeAsync(
+                entity.Key,
+                entity.TenantId,
+                entity.AgentId,
+                entity.UserId,
+                entity.ActivationName,
+                entity.Id);
+            if (existsForScope)
+                return ServiceResult<SecretVaultGetResponse>.Conflict("A secret with this key already exists for the same scope");
 
             entity.UpdatedAt = DateTime.UtcNow;
             entity.UpdatedBy = actorUserId;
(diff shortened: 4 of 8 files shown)


xiana-agent (changes requested):
## PR Review Report

**PR:** Refactor Secret Vault endpoints to use key-based access + tenant validation and scope uniqueness enforcement
**Author:** Kavish Nanayakkara <KavishN@99x.io>
**Files Changed:** 8 | **+305** / **-154**

---

### Summary

This PR refactors the Secret Vault API from ID-based to key+scope-based addressing (PUT and DELETE), adds tenant existence validation via `ITenantCacheService` on create/update/fetch operations, and introduces composite key+scope uniqueness enforcement at the repository and service layers. The architectural direction is sound and the uniqueness enforcement via `ExistsByKeyAndScopeAsync` is well-structured. However, there is a critical compilation-breaking bug: both update request DTOs are missing the `Key` property that the endpoint code tries to access.

---

- [ ] `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs:128` — `SecretVaultUpdateRequest` does not declare a `Key` property, but `request.Key` is accessed on lines 128 and 164. This is a compilation error that will prevent the project from building.

  ```csharp
  // Current (broken) — SecretVaultUpdateRequest has no Key property
  public class SecretVaultUpdateRequest
  {
      public string? Value { get; set; }
      public string? TenantId { get; set; }
      // ... no Key field
  }

  // Fix — add Key to the update request DTO
  public class SecretVaultUpdateRequest
  {
      public string Key { get; set; } = null!;
      public string? Value { get; set; }
      public string? TenantId { get; set; }
      public string? AgentId { get; set; }
      public string? UserId { get; set; }
      public string? ActivationName { get; set; }
      public object? AdditionalData { get; set; }
  }
  ```

- [ ] `XiansAi.Server.Src/Features/AgentApi/Endpoints/SecretVaultEndpoints.cs:141` — Same issue: `AgentSecretVaultUpdateRequest` is missing `Key`, but `request.Key` is accessed on lines 141 and 177.

  ```csharp
  // Fix — add Key to the agent update request DTO
  public class AgentSecretVaultUpdateRequest
  {
      public string Key { get; set; } = null!;
      public string? Value { get; set; }
      // ... rest of fields
  }
  ```

---

### Warnings (Should Fix)

- [ ] `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs:39` — The Admin API tenant validation guard condition is stricter than the Agent API equivalent. The Admin API only validates tenant existence when `request.TenantId` is non-empty **and** `effectiveTenantId` is non-empty. The Agent API (line 40 in `SecretVaultEndpoints.cs`) validates whenever `effectiveTenantId` is non-empty. This asymmetry means an Admin caller whose scope resolves to a non-null `effectiveTenantId` (e.g. via role-based scope injection that does not require explicit `request.TenantId`) would bypass the tenant existence check. Consider aligning both APIs to check `!string.IsNullOrWhiteSpace(effectiveTenantId)` as the sole guard.

- [ ] `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs:312` — The normalization expressions in `UpdateByKeyAsync` are redundant. `input.TenantId == null ? null : (string.IsNullOrWhiteSpace(input.TenantId) ? null : input.TenantId)` is equivalent to `string.IsNullOrWhiteSpace(input.TenantId) ? null : input.TenantId` because `null` is already whitespace-falsy through `IsNullOrWhiteSpace`. The extra null-check on the outer ternary is dead code and adds confusion. Use the same one-liner pattern already applied in `CreateAsync`.

- [ ] `XiansAi.Server.Src/Shared/Repositories/SecretVaultRepository.cs:101` — `ExistsByKeyAndScopeAsync` uses `CountDocumentsAsync` which scans all matching documents to return a count. Since only a boolean is needed, prefer `Find(filter).Limit(1).AnyAsync()` (or `CountDocumentsAsync(filter, new CountOptions { Limit = 1 })`) so MongoDB can short-circuit after finding the first match, especially important if the collection grows large.

- [ ] `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs:306` — `UpdateByKeyAsync` delegates to `UpdateAsync(entity.Id, ...)` which performs a second repository lookup (`GetByIdAsync`) for the entity that was just fetched. This is an unnecessary extra round-trip to MongoDB. The entity is already in scope; the update logic could be applied directly, or `UpdateAsync` could be refactored to accept an entity instead of an id.

---

- [ ] `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs:61` — The OpenAPI description on `CreateSecret` still says "Key must be unique" which is now outdated — key uniqueness is scoped to the key+scope tuple, not global. Update to "Key must be unique within its scope."

- [ ] `XiansAi.Server.Src/Shared/Repositories/SecretVaultRepository.cs:90` — `ExistsByKeyAndScopeAsync` silently returns `false` on exception (line 107), which means a database failure during a uniqueness check would allow a duplicate to be created. Consider re-throwing or returning a distinct error result to surface infrastructure failures rather than masking them as "no conflict."

- [ ] `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs` — No automated unit or integration tests were added for any of the new behaviours: tenant validation returning 404, composite uniqueness enforcement on create and update, key-based update/delete with not-found handling, or the scope asymmetry between Admin and Agent APIs. The existing test coverage pattern in the project uses xUnit integration tests; the new paths should follow suit.

- [ ] `XiansAi.Server.Src/docs/SECRET_VAULT.md:1` — Top-level heading changed from `# Secret Vault` to `## Secret Vault`, making it a second-level heading as the document root. This may render incorrectly when the document is viewed standalone or included in a documentation site. Restore `# Secret Vault`.

---

### Review Details

#### Code Quality

The refactoring direction is clean: replacing ID-based mutation endpoints with key+scope-based ones eliminates the need for callers to retain internal MongoDB IDs, which is a significant usability and API design improvement. The `BuildKeyAndScopeFilter` private helper correctly extracts duplicated filter construction into a single location. The `UpdateByKeyAsync` / `DeleteByKeyAsync` bridge methods follow a reasonable delegation pattern. The critical problem is the two missing `Key` fields on the update request DTOs — this is a build-breaking omission that must have been introduced when renaming the update DTO properties and would be caught immediately by the compiler.

#### Security

The addition of tenant existence validation before mutating or reading secrets is a good defensive measure. However, the behavioural asymmetry between Admin and Agent API tenant guards (described in Warnings above) could allow an Admin-scoped operation to bypass the check under certain scope-resolution paths. Secret values are still encrypted at rest and decryption keys are properly read from configuration. No secrets or tokens are introduced in this diff. The `DELETE` endpoint now accepts all scope parameters as plain query strings; these are not sensitive by themselves (keys are identifiers, not secrets), so no additional concern there.

#### Test Coverage

No automated tests were added or modified as part of this PR. Other feature areas in this project (e.g. `AdminTenantEndpointsTests.cs`, `ApiKeyEndpointsTests.cs`) have corresponding integration test classes. The new tenant-validation logic, uniqueness enforcement, and key-based CRUD paths all represent non-trivial business rules that should be covered. The `.http` test files are updated correctly for manual verification, but they do not substitute for automated regression coverage.

#### Performance

The double MongoDB round-trip in `UpdateByKeyAsync` (fetch by key+scope, then fetch by ID again inside `UpdateAsync`) is unnecessary overhead and should be addressed. The use of `CountDocumentsAsync` without a document limit in `ExistsByKeyAndScopeAsync` is a minor inefficiency that will become more noticeable at scale. The `BuildKeyAndScopeFilter` helper and the refactored `FindForAccessAsync` correctly share the same filter path, which is good for maintainability.

---

### Files Reviewed

| `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs` | +56/-41 | High | Missing `Key` on update DTO; tenant guard asymmetry |
| `XiansAi.Server.Src/Features/AgentApi/Endpoints/SecretVaultEndpoints.cs` | +68/-42 | High | Missing `Key` on update DTO |
| `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs` | +51/-16 | High | New key-based methods; double round-trip in UpdateByKeyAsync |
| `XiansAi.Server.Src/Shared/Repositories/SecretVaultRepository.cs` | +48/-27 | Medium | New ExistsByKeyAndScopeAsync; CountDocuments without limit |
| `XiansAi.Server.Src/docs/SECRET_VAULT.md` | +41/-17 | Low | Good documentation; heading level regression |
| `XiansAi.Server.Src/docs/SECRET_VAULT_VALIDATION.md` | +4/-7 | Low | Accurately reflects the API changes |
| `XiansAi.Server.Tests/http/AdminApi/admin-secret-vault-endpoints.http` | +6/-13 | Low | Correctly updated for key-based operations |
| `XiansAi.Server.Tests/http/AgentApi/agent-secret-vault-endpoints.http` | +6/-13 | Low | Correctly updated for key-based operations |

xiana-agent (changes requested):
## PR Review Report

**PR:** Secret Vault Key + Scope Only & Tenant Existence Validation and also Ensuring scope level key uniqueness
**Author:** Kavish Nanayakkara <KavishN@99x.io>
**Files Changed:** 8 | **+308** / **-155**

---

### Summary

This PR replaces id-based secret vault operations (GET/PUT/DELETE by `/{id}`) with key+scope-based operations across both the Admin and Agent APIs, adds tenant existence validation via `ITenantCacheService`, and enforces composite uniqueness on (key, tenantId, agentId, userId, activationName). The refactor is architecturally sound and well-motivated. However, there are two notable correctness issues and several important gaps to address before merge.

---

- [ ] `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs:312-315` — **Redundant null-check makes the normalization a no-op.** The expression `input.TenantId == null ? null : (string.IsNullOrWhiteSpace(input.TenantId) ? null : input.TenantId)` is logically identical to `string.IsNullOrWhiteSpace(input.TenantId) ? null : input.TenantId` because `IsNullOrWhiteSpace` already handles the null case. The extra outer null guard is dead code and creates a misleading asymmetry with the normalization logic in `CreateAsync` (lines 110-113). This is fine on its own, but the same pattern in `UpdateByKeyAsync` normalises the lookup scope — however `FindForAccessAsync` then receives the same normalised values that `UpdateAsync` will use internally for its own `ExistsByKeyAndScopeAsync` check. If the update body carries a scope different from the lookup scope, the uniqueness check in `UpdateAsync` will use the *new* (updated) scope on the entity (which may differ from the lookup scope), meaning the find and the conflict-check are operating on different scope tuples. This TOCTOU window is not exploitable under normal use, but the double fetch (first `FindForAccessAsync`, then `GetByIdAsync` inside `UpdateAsync`) means the service performs **two separate MongoDB round-trips** to retrieve the same document. This is wasteful and carries a TOCTOU window if the document is mutated between calls.

  ```csharp
  // Current — two round-trips, TOCTOU window
  public async Task<ServiceResult<SecretVaultGetResponse>> UpdateByKeyAsync(...)
  {
      var entity = await _repository.FindForAccessAsync(key, ...);   // round-trip 1
      if (entity == null) return NotFound(...);
      return await UpdateAsync(entity.Id, input, actorUserId);       // GetByIdAsync inside = round-trip 2
  }

  // Fix — pass the already-fetched entity directly, or add an overload that accepts the entity
  public async Task<ServiceResult<SecretVaultGetResponse>> UpdateByKeyAsync(...)
  {
      var entity = await _repository.FindForAccessAsync(key, ...);
      if (entity == null) return NotFound(...);
      return await UpdateEntityAsync(entity, input, actorUserId);    // skip second GetByIdAsync
  }
  ```

- [ ] `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs:39` — **Inconsistent tenant-validation guard between Admin and Agent APIs.** The Admin API checks `!string.IsNullOrWhiteSpace(request.TenantId) && !string.IsNullOrWhiteSpace(effectiveTenantId)` (both conditions required), whereas the Agent API checks only `!string.IsNullOrWhiteSpace(effectiveTenantId)`. This means an Admin API caller whose `TenantId` is null but whose effective tenant is resolved from the JWT/context (a common scenario for non-SysAdmin admins) will **skip tenant existence validation**, while the Agent API performs it. This is a security inconsistency — either both APIs should validate whenever an effective tenant is resolved, or the intentional difference should be explicitly documented and justified.

  ```csharp
  // Admin API — current (potentially skips validation for context-resolved tenants)
  if (!string.IsNullOrWhiteSpace(request.TenantId) && !string.IsNullOrWhiteSpace(effectiveTenantId))

  // Fix — validate whenever an effective tenant is resolved (matches Agent API behaviour)
  if (!string.IsNullOrWhiteSpace(effectiveTenantId))
  ```

---

### Warnings (Should Fix)

- [ ] `XiansAi.Server.Src/Shared/Repositories/SecretVaultRepository.cs:101` — **`CountDocumentsAsync` is heavier than `Find(...).Limit(1).AnyAsync()`** for an existence check. MongoDB must count all matching documents unless the planner uses an index-only plan. Using `AnyAsync` on a `Find(...).Limit(1)` avoids a full scan when more than one document matches and more clearly communicates intent.

  ```csharp
  // Current
  var count = await _collection.CountDocumentsAsync(filter);
  return count > 0;

  // Preferred
  return await _collection.Find(filter).Limit(1).AnyAsync();
  ```

- [ ] `XiansAi.Server.Src/Shared/Repositories/SecretVaultRepository.cs` — **No composite index documented or enforced for (key, tenantId, agentId, userId, activationName).** `BuildKeyAndScopeFilter` is now the hot path for fetch, update, delete, and both existence checks. Without a compound index on these five fields, every call does a collection scan. At minimum a comment, migration note, or `EnsureIndex` call at startup should be present.

- [ ] `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs:306-323` — **`UpdateByKeyAsync` delegates to `UpdateAsync(entity.Id, ...)` which calls `GetByIdAsync(id)` again.** This is an extra unnecessary database read on every update by key. This ties back to the critical issue above, but even without a TOCTOU concern the double fetch is a performance regression for all key-based update calls.

- [ ] `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs:121-165` — **`PUT /` (update) accepts a request body but HTTP DELETE `DELETE /` (line 170-201) accepts only query parameters, including `key`.** Sending a secret key as a plain query parameter means the key appears in server access logs, browser history, and HTTP proxies. While this may be acceptable in some internal environments, sensitive identifiers should generally be sent in request headers or body. The existing `GET /fetch` has the same pattern; but for a mutating DELETE this is at minimum worth a documented decision.

- [ ] `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs` and `SecretVaultEndpoints.cs` — **`SecretVaultCreateRequest.Key` and `SecretVaultUpdateRequest.Key` are declared as `= null!` (non-nullable reference with null-forgiving operator)** but no `[Required]` data annotation is present. Model binding in minimal APIs does not enforce non-nullability at runtime; a missing `key` field in the JSON body will deserialize to `null` without a 400 response from the framework — the validation is only caught downstream in the service. Adding `[Required]` (or `[BindRequired]`) returns a clear 400 before hitting service logic.

---

- [ ] `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs:312-315` — The normalization pattern `x == null ? null : (string.IsNullOrWhiteSpace(x) ? null : x)` is duplicated four times in `UpdateByKeyAsync`. Extract to a small inline helper `NormalizeScope(string? v) => string.IsNullOrWhiteSpace(v) ? null : v` (already implicitly done in `CreateAsync`) to DRY this up.

- [ ] `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs` — The `ExistsByKeyAsync` method on `ISecretVaultRepository` (line 11 of the repository interface) is now only used by legacy code paths. If it is no longer called by any live endpoint, consider removing it or marking it `[Obsolete]` to reduce confusion about which existence-check method to use.

- [ ] `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs` and `SecretVaultEndpoints.cs` — The `SecretVaultUpdateRequest`/`AgentSecretVaultUpdateRequest` classes are defined in the endpoint files (alongside `CreateRequest`). Since `Key` is now a first-class field on the update request, consider unifying the Admin and Agent request DTOs into shared types in the `Shared` layer to avoid the current parallel-class maintenance burden.

- [ ] `XiansAi.Server.Src/docs/SECRET_VAULT.md:1` — The top-level heading was changed from `# Secret Vault` to `## Secret Vault`, making it a second-level heading. If this is the root document it should remain an H1.

---

### Review Details

#### Code Quality

The refactor is clean and well-structured. `BuildKeyAndScopeFilter` is a good extraction — the duplication between `FindForAccessAsync` and the old inline filter construction is removed. The `UpdateByKeyAsync` / `DeleteByKeyAsync` service methods follow a consistent pattern. The `UpdateRequest.Key = null!` pattern is the main stylistic concern; the absence of `[Required]` means the non-null enforcement is implicit and invisible to API consumers and OpenAPI tooling. The normalization helper for scope values is duplicated between `CreateAsync` and `UpdateByKeyAsync` and would benefit from extraction.

#### Security

The most significant security finding is the inconsistent tenant-validation guard between Admin and Agent APIs (critical issue #2 above). The key-in-query-string pattern for DELETE is a low-severity information-disclosure risk. No hardcoded secrets or injection vectors were found. Scope enforcement via `TryResolveScope` is applied consistently before any data access on all mutating paths. The `ExistsByKeyAndScopeAsync` check-then-insert pattern is not atomic (no MongoDB transaction or unique index), so a race condition between two concurrent creates with identical key+scope is theoretically possible — this is a pre-existing pattern, but worth noting given the explicit uniqueness guarantee now documented.

#### Test Coverage

No automated tests were added or modified for the new key-based endpoints (`UpdateByKeyAsync`, `DeleteByKeyAsync`), the new `ExistsByKeyAndScopeAsync` repository method, or the tenant-existence validation paths. The test suite contains integration tests for other Admin and Agent API endpoints but none for Secret Vault. The `.http` test files are helpful for manual testing but do not substitute for automated coverage. The following scenarios have no test coverage: (1) create with duplicate key+scope returns 409, (2) update by key where secret does not exist returns 404, (3) delete by key with mismatched scope returns 404, (4) tenant not found returns 404, (5) Agent API vs Admin API tenant-validation asymmetry.

#### Performance

Two MongoDB round-trips per `UpdateByKeyAsync` call (see critical issue #1 and warning #3). `CountDocumentsAsync` used for existence checks instead of the more efficient `Find(...).Limit(1).AnyAsync()`. No compound index on the (key, scope) tuple is documented, which means every fetch/update/delete/exists call will scan the entire `secret_vault` collection as the collection grows.

---

### Files Reviewed

| `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs` | +75/-57 | High | Auth/scope logic, tenant validation asymmetry |
| `XiansAi.Server.Src/Features/AgentApi/Endpoints/SecretVaultEndpoints.cs` | +81/-61 | High | Certificate-auth path, tenant validation |
| `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs` | +51/-4 | High | Core business logic, double fetch, normalization |
| `XiansAi.Server.Src/Shared/Repositories/SecretVaultRepository.cs` | +53/-26 | Medium | MongoDB filter extraction, existence check |
| `XiansAi.Server.Src/docs/SECRET_VAULT.md` | +43/-14 | Low | Documentation, H1 heading changed to H2 |
| `XiansAi.Server.Src/docs/SECRET_VAULT_VALIDATION.md` | +4/-7 | Low | Documentation update |
| `XiansAi.Server.Tests/http/AdminApi/admin-secret-vault-endpoints.http` | +5/-14 | Low | Manual test file |
| `XiansAi.Server.Tests/http/AgentApi/agent-secret-vault-endpoints.http` | +8/-15 | Low | Manual test file |

hasith:
**Review in progress** — A comprehensive automated review is being conducted across code quality, security, tests, and performance dimensions. Report will be posted shortly.

hasith (commented):
## Automated PR Review — Secret Vault Key + Scope Only & Tenant Existence Validation

**Verdict: APPROVE WITH REQUESTS** — The implementation is architecturally sound and the feature logic is correct. Several issues ranging from critical bugs to design inconsistencies require attention before merge.

---

### Summary

| Dimension | Finding Count | Highest Severity |
|-----------|--------------|-----------------|
| Code Quality | 5 | WARNING |
| Security | 3 | WARNING |
| Tests | 1 | CRITICAL |
| Performance | 2 | SUGGESTION |

**Files reviewed:**
- `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs`
- `XiansAi.Server.Src/Features/AgentApi/Endpoints/SecretVaultEndpoints.cs`
- `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs`
- `XiansAi.Server.Src/Shared/Repositories/SecretVaultRepository.cs`
- `XiansAi.Server.Src/Shared/Auth/SecretVaultScopeEnforcement.cs`
- `XiansAi.Server.Src/Shared/Services/TenantCacheService.cs`

---

## CRITICAL

### [CRITICAL-1] `GetByTenantIdAsync` throws `ValidationException` for malformed tenant IDs — unhandled in all endpoints

**File:** `SecretVaultEndpoints.cs` lines 41–46, 121–126, 162–167; `AdminSecretVaultEndpoints.cs` lines 39–46, 147–154

`TenantCacheService.GetByTenantIdAsync` calls `Tenant.SanitizeAndValidateTenantId` internally, which throws `System.ComponentModel.DataAnnotations.ValidationException` when the tenant ID is null/empty/malformed. None of the six endpoint call-sites that invoke `tenantCacheService.GetByTenantIdAsync` catch this exception, so a malformed `tenantId` in the request body or query string will cause an unhandled exception — producing a 500 response instead of the intended 400 or 404.

**Affected code (Agent API create — same pattern repeated in 5 more places):**
```csharp
// SecretVaultEndpoints.cs ~line 39-46
if (!string.IsNullOrWhiteSpace(effectiveTenantId))
{
    var tenant = await tenantCacheService.GetByTenantIdAsync(effectiveTenantId, cancellationToken);
    if (tenant == null)
        return ServiceResult<SecretVaultGetResponse>.NotFound("Tenant not found").ToHttpResult();
}
```

**Fix:** Wrap each call in try/catch, or add a pre-validation step before invoking the cache service:
```csharp
if (!string.IsNullOrWhiteSpace(effectiveTenantId))
{
    try
    {
        var tenant = await tenantCacheService.GetByTenantIdAsync(effectiveTenantId, cancellationToken);
        if (tenant == null)
            return ServiceResult<SecretVaultGetResponse>.NotFound("Tenant not found").ToHttpResult();
    }
    catch (System.ComponentModel.DataAnnotations.ValidationException)
    {
        return ServiceResult<SecretVaultGetResponse>.BadRequest("Invalid tenant ID format").ToHttpResult();
    }
}
```
Alternatively, add a format pre-check before calling the service.

---

## WARNINGS

### [WARN-1] Race condition (TOCTOU) in scope-level key uniqueness — no database-level unique compound index

**File:** `SecretVaultService.cs` lines 115–122 (create), lines 233–241 (update); `SecretVaultRepository.cs` line 90

Uniqueness is enforced by an application-level check (`ExistsByKeyAndScopeAsync`) followed by a separate insert/update. Under concurrent requests, two simultaneous creates with the same key+scope can both pass the existence check before either completes the write, resulting in duplicate documents. The `mongodb-indexes.yaml` file has **no compound unique index** defined for the `secret_vault` collection.

**Current approach (check-then-act, not atomic):**
```csharp
var exists = await _repository.ExistsByKeyAndScopeAsync(input.Key, ...);
if (exists) return Conflict(...);
// <-- another request can sneak in here -->
await _repository.CreateAsync(entity);
```

**Fix:** Add a compound unique sparse index to `mongodb-indexes.yaml`:
```yaml
secret_vault:
  - name: key_tenant_agent_user_activation_unique
    keys:
      key: asc
      tenant_id: asc
      agent_id: asc
      user_id: asc
      activation_name: asc
    unique: true
    sparse: true
```
Then handle `MongoWriteException` (duplicate key error code 11000) in `CreateAsync` and `UpdateAsync` as an additional conflict guard. The application-level check can remain as a fast-path to return a friendly error before hitting the DB.

---

### [WARN-2] Tenant existence check asymmetry between Agent API and Admin API

**File:** `SecretVaultEndpoints.cs` line 40 vs. `AdminSecretVaultEndpoints.cs` line 39

In the Agent API, the tenant existence check triggers whenever `effectiveTenantId` is non-empty (i.e., any resolved tenant scope). In the Admin API, the check additionally requires that `request.TenantId` is non-empty — meaning a SysAdmin who omits `tenantId` from the body but whose `effectiveTenantId` is somehow resolved would bypass the tenant existence check in the Admin API.

**Agent API (correct — validates all resolved tenant IDs):**
```csharp
if (!string.IsNullOrWhiteSpace(effectiveTenantId))
{
    var tenant = await tenantCacheService.GetByTenantIdAsync(effectiveTenantId, cancellationToken);
```

**Admin API (narrower check — only validates when request body also has tenantId):**
```csharp
if (!string.IsNullOrWhiteSpace(request.TenantId) && !string.IsNullOrWhiteSpace(effectiveTenantId))
{
    var tenant = await tenantCacheService.GetByTenantIdAsync(effectiveTenantId, cancellationToken);
```

**Fix:** Align Admin API to match Agent API semantics:
```csharp
if (!string.IsNullOrWhiteSpace(effectiveTenantId))
{
    var tenant = await tenantCacheService.GetByTenantIdAsync(effectiveTenantId, cancellationToken);
    if (tenant == null)
        return ServiceResult<SecretVaultGetResponse>.NotFound("Tenant not found").ToHttpResult();
}
```
Also note the Admin API fetch endpoint (`GET /fetch`) has no tenant validation at all, unlike its Agent API counterpart. This should be added for consistency.

---

### [WARN-3] `UpdateByKeyAsync` performs a double DB round-trip through `FindForAccessAsync` + `UpdateAsync` (which calls `GetByIdAsync` again)

**File:** `SecretVaultService.cs` lines 306–323

`UpdateByKeyAsync` finds the entity by key+scope (one DB call), then delegates to `UpdateAsync(entity.Id, ...)`, which calls `GetByIdAsync(id)` again (a second DB call) to reload the same entity. This is redundant and introduces a window where the entity could be deleted between the two reads.

```csharp
public async Task<ServiceResult<SecretVaultGetResponse>> UpdateByKeyAsync(...)
{
    var entity = await _repository.FindForAccessAsync(key, ...);  // DB call #1
    if (entity == null) return NotFound(...);

    return await UpdateAsync(entity.Id, input, actorUserId);  // calls GetByIdAsync internally → DB call #2
}
```

**Fix:** Refactor `UpdateAsync` to accept the already-loaded entity, or inline the update logic directly in `UpdateByKeyAsync` to avoid reloading. At minimum, document that this is intentional if the double-load is desired for freshness.

---

### [WARN-4] Orphaned interface methods `ExistsByKeyAsync` and `GetByKeyAsync` on `ISecretVaultRepository`

**File:** `SecretVaultRepository.cs` lines 10–11

The methods `GetByKeyAsync(string key)` and `ExistsByKeyAsync(string key)` remain on the `ISecretVaultRepository` interface and have concrete implementations, but are no longer called anywhere in the service layer (the service now uses `ExistsByKeyAndScopeAsync` and `FindForAccessAsync`). These should be removed to keep the interface clean and avoid confusion about which uniqueness check to use.

**Fix:** Remove `ExistsByKeyAsync(string key)` and `GetByKeyAsync(string key)` from the interface and implementation, as `FindForAccessAsync` and `ExistsByKeyAndScopeAsync` fully supersede them. Confirm no external consumers reference these methods via the interface.

---

### [WARN-5] `SecretVaultUpdateRequest.Key` and `AgentSecretVaultUpdateRequest.Key` initialized as `null!` but not annotated as required

**File:** `AdminSecretVaultEndpoints.cs` line 220; `SecretVaultEndpoints.cs` line 232

Both update request classes declare `Key` with `= null!` (null-forgiving initializer), which suppresses nullable warnings but means a request body that omits the `key` field will bind as `null` at runtime. The endpoint immediately checks `IsNullOrWhiteSpace(request.Key)` and returns 400, but the model allows null silently. This differs from `SecretVaultCreateRequest.Key` which has no null-forgiving assignment and will fail model binding, providing a clearer framework-level error.

**Current:**
```csharp
public string Key { get; set; } = null!;
```

**Preferred:** Add `[Required]` attribute and/or remove the `= null!` to let the model binder surface a 400 before the handler runs, consistent with `SecretVaultCreateRequest`:
```csharp
[Required]
public string Key { get; set; } = string.Empty;
```

---

## SECURITY

### [SEC-1] Tenant validation not applied on Admin API `GET /fetch` endpoint

**File:** `AdminSecretVaultEndpoints.cs` lines 91–119

The Admin API's fetch-by-key endpoint resolves scope and calls `FetchByKeyAsync`, but unlike the Agent API's equivalent endpoint (which added tenant existence validation in this PR), the Admin API fetch endpoint has no tenant existence check. A caller can fetch secrets using a tenantId that references a deleted or non-existent tenant.

**Fix:** Add the same tenant existence guard to Admin API fetch, mirroring what was added to Agent API fetch:
```csharp
if (!string.IsNullOrWhiteSpace(effectiveTenantId))
{
    var tenant = await tenantCacheService.GetByTenantIdAsync(effectiveTenantId, cancellationToken);
    if (tenant == null)
        return ServiceResult<SecretVaultFetchResponse?>.NotFound("Tenant not found").ToHttpResult();
}
```
Note this also requires adding `ITenantCacheService` and `CancellationToken` to the fetch handler's parameters.

---

### [SEC-2] Admin API delete has no tenant existence validation

**File:** `AdminSecretVaultEndpoints.cs` lines 170–201

The delete endpoint for Admin API resolves scope correctly via `TryResolveScope` and calls `DeleteByKeyAsync`, but does not validate tenant existence (unlike create and update in the same file). This is inconsistent and allows deletion of secrets tied to a scope that references a non-existent tenant.

**Fix:** Add tenant validation in the Admin API delete handler, consistent with create and update.

---

### [SEC-3] `ExistsByKeyAndScopeAsync` silently returns `false` on database error — create proceeds as if no duplicate exists

**File:** `SecretVaultRepository.cs` lines 90–108

When `CountDocumentsAsync` throws an exception, `ExistsByKeyAndScopeAsync` catches it and returns `false`. This means on a transient DB error during the existence check, the service layer treats it as "no duplicate found" and proceeds to insert, potentially creating a duplicate if the first check failed mid-flight. The same silent `false` return exists in `FindForAccessAsync` (returns `null`).

```csharp
catch (Exception ex)
{
    _logger.LogError(ex, "Error checking existence for key {Key} and scope", key);
    return false;  // Incorrect: should propagate the error upward
}
```

**Fix:** Re-throw the exception (or throw a domain-specific exception) so the service layer can return 500 instead of proceeding optimistically:
```csharp
catch (Exception ex)
{
    _logger.LogError(ex, "Error checking existence for key {Key} and scope", key);
    throw; // Let the service layer handle and return InternalServerError
}
```

---

## TESTS

### [TEST-1] No automated tests for any new functionality introduced in this PR

**File:** `XiansAi.Server.Tests/` — no matching test files found

This PR introduces significant behavioral changes:
- `UpdateByKeyAsync` and `DeleteByKeyAsync` service methods
- Scope-level key uniqueness enforcement via `ExistsByKeyAndScopeAsync`
- Tenant existence validation on create, update, and fetch across both Agent and Admin APIs
- Removal of the id-based GET/PUT/DELETE endpoints (breaking changes)

None of these are covered by unit tests or integration tests. The existing test suite has no `*SecretVault*` test files at all.

**Required test cases:**
1. Create with duplicate key+scope returns 409
2. Create with same key but different scope succeeds
3. Create with non-existent tenantId returns 404
4. Update by key+scope finds correct secret and updates it
5. Delete by key+scope removes correct secret
6. Update with a key+scope collision (excluding self) returns 409
7. Agent API: tenant validation on create/update/fetch gates on `effectiveTenantId`
8. Admin API: `request.TenantId` gate condition for tenant validation
9. Requests to removed `GET /{id}`, `PUT /{id}`, `DELETE /{id}` return 404/405

At minimum, unit tests for `SecretVaultService.CreateAsync`, `UpdateByKeyAsync`, `DeleteByKeyAsync`, and `ExistsByKeyAndScopeAsync` should be added using a mocked repository.

---

## SUGGESTIONS

### [SUGG-1] Redundant null-coalescing in `UpdateByKeyAsync` normalization

**File:** `SecretVaultService.cs` lines 312–315

The normalization expressions are verbose and redundant — the outer `input.X == null ? null :` is always `null` when the inner `string.IsNullOrWhiteSpace` is also `null`, making both branches reduce to the same thing.

**Current (verbose):**
```csharp
var lookupTenantId = input.TenantId == null ? null : (string.IsNullOrWhiteSpace(input.TenantId) ? null : input.TenantId);
```

**Simplified:**
```csharp
var lookupTenantId = string.IsNullOrWhiteSpace(input.TenantId) ? null : input.TenantId;
```
This already matches the pattern used in `CreateAsync` (lines 110–113) and is simpler to read.

---

### [SUGG-2] `SecretVault` model comment still says "Key is unique in the collection" — now incorrect

**File:** `SecretVaultRepository.cs` line 7 (XML doc on `SecretVault` model)

The `SecretVault` class XML comment says "Key (secret name) is unique in the collection", but uniqueness is now enforced on the key+scope tuple, not key alone. This misleads future contributors.

**Fix:**
```csharp
/// Secret vault document. The combination of Key + scope tuple (TenantId, AgentId, UserId, ActivationName)
/// is unique. The same Key may appear in multiple documents as long as the scope differs.
```

---

## Positive Observations

- The `BuildKeyAndScopeFilter` private helper is a clean refactor — good extraction of the repeated filter-building logic that was previously duplicated between `FindForAccessAsync` and `ExistsByKeyAndScopeAsync`.
- The `UpdateAsync` method correctly uses `excludeId` when checking uniqueness after mutating scope fields, avoiding false conflict detection on the document being updated.
- `TenantCacheService` uses a semaphore-per-key pattern and caches null results with a shorter TTL — the tenant validation calls are not going to add significant overhead.
- The scope normalization (whitespace → null) is correctly applied before writes, ensuring consistent storage semantics.
- Removal of id-based endpoints is correct; key+scope is a more stable and appropriate identifier for external consumers.

---

*Review generated by automated PR review agent — 2026-04-19*

hasith:
**Review in Progress**

A comprehensive automated review has been started for this PR. Analysis is being performed across code quality, security, test coverage, and performance. Results will be posted shortly.

hasith:
## Pull Request Review — #385
**Secret Vault Key + Scope Only & Tenant Existence Validation**

**Verdict: REQUEST CHANGES**

---

### Summary

This PR delivers a meaningful improvement: secret operations now use `key + scope` as the external identity instead of opaque MongoDB `_id` values, tenant existence is validated before persisting secrets, and uniqueness enforcement is promoted to the composite `(key, tenantId, agentId, userId, activationName)` tuple. The direction is correct. However, several issues must be addressed before merging.

---

### CRITICAL

#### 1. No database-level unique index — race condition on concurrent creates/updates

**Files:** `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs` (lines ~115, ~233), `XiansAi.Server.Src/mongodb-indexes.yaml`

The service enforces key+scope uniqueness by running `ExistsByKeyAndScopeAsync` (a count query) and then, on success, calling `CreateAsync` or `UpdateAsync`. This is a check-then-act pattern with no transaction and no unique index to back it up. Two concurrent requests with the same key and scope can both pass the count check simultaneously and each proceed to insert/update, resulting in duplicate documents. MongoDB's `secret_vault` collection has **no** composite unique index defined in `mongodb-indexes.yaml`, so there is no database-level safety net.

**Required fix:** Add a sparse compound unique index to `mongodb-indexes.yaml`:

```yaml
secret_vault:
  - name: key_scope_unique
    keys:
      key: asc
      tenant_id: asc
      agent_id: asc
      user_id: asc
      activation_name: asc
    unique: true
    sparse: true
```

And handle the `MongoWriteException` / `MongoBulkWriteException` with `Category == DuplicateKey` in `SecretVaultRepository.CreateAsync` (currently it rethrows any exception, so the service would return a 500 instead of a 409).

---

#### 2. `SecretVaultService.CreateAsync` still calls the old `ExistsByKeyAsync` (no scope) — the PR branch has a divergence

**File:** `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs` line ~107

The PR diff shows the service being updated to call `ExistsByKeyAndScopeAsync`, but the local `main`-branch version of `SecretVaultService.cs` (which is what is checked in) still calls `ExistsByKeyAsync(input.Key, input.TenantId)` — the legacy two-argument version — rather than `ExistsByKeyAndScopeAsync`. This means the full-scope uniqueness check for create is **missing** on `main`, and the PR's primary goal is not achieved. Confirm that the PR branch version is the intended source of truth and that it merges cleanly.

---

### WARNINGS

#### 3. Inconsistent tenant-existence guard between Admin API and Agent API

**Files:**
- `AdminSecretVaultEndpoints.cs` (Create and Update): tenant check only fires when `request.TenantId` is **explicitly** non-empty _and_ `effectiveTenantId` is non-empty.
- `SecretVaultEndpoints.cs` (Agent API Create, Update, and Fetch): tenant check fires whenever `effectiveTenantId` is non-empty (no requirement that the caller explicitly specified a tenantId).

For the Admin API, a SysAdmin whose certificate or context already resolves a `tenantId` without the caller setting `request.TenantId` explicitly will bypass the tenant existence check. The Agent API is stricter (it always validates when `effectiveTenantId` is resolved). Pick one consistent rule and apply it in both files.

**Example Admin API gap (Create endpoint, line ~38):**
```csharp
// Current — misses SysAdmin flow where tenantId is resolved from context, not from request
if (!string.IsNullOrWhiteSpace(request.TenantId) && !string.IsNullOrWhiteSpace(effectiveTenantId))

// Should be consistent with Agent API:
if (!string.IsNullOrWhiteSpace(effectiveTenantId))
```

---

#### 4. `DELETE` endpoint does not validate tenant existence before attempting delete

**Files:** `AdminSecretVaultEndpoints.cs` (Delete handler), `SecretVaultEndpoints.cs` (Agent API Delete handler)

Both Delete endpoints resolve scope and call `DeleteByKeyAsync` directly, with no check that the referenced tenant exists. The PR description says tenant existence is validated on create, update, and fetch, but not on delete. A delete for a non-existent tenant will return 404 ("Secret not found") — which is acceptable — but it would be more explicit and consistent to return 404 ("Tenant not found") before attempting the lookup when a tenantId is provided. This is a policy decision, but should be documented if left inconsistent.

---

#### 5. `UpdateByKeyAsync` — redundant double-null normalization pattern

**File:** `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs` lines ~318-321

```csharp
var lookupTenantId = input.TenantId == null ? null : (string.IsNullOrWhiteSpace(input.TenantId) ? null : input.TenantId);
```

The outer `null` check is redundant because `string.IsNullOrWhiteSpace(null)` already returns `true`. The simplified form is cleaner:

```csharp
var lookupTenantId = string.IsNullOrWhiteSpace(input.TenantId) ? null : input.TenantId;
```

This pattern appears four times in `UpdateByKeyAsync`. More importantly, the `input` object passed to `FindForAccessAsync` for lookup and then the same `input` passed to `UpdateAsync` contain the **caller-supplied scope values**, which may differ from the document's stored scope if the caller wants to change scope on update. This means the lookup uses the new target scope, not the current document scope — the lookup will fail to find the secret unless the caller passes the current scope to identify it. The API contract for update-by-key needs to distinguish between "the scope used to identify the secret" and "the new scope to set." Currently these are the same object, which is ambiguous.

---

#### 6. Old `GetByKeyAsync` and `ExistsByKeyAsync` (no scope) remain in the interface but are unused by the new code paths

**File:** `XiansAi.Server.Src/Shared/Repositories/SecretVaultRepository.cs` lines 10-11

The interface still declares `GetByKeyAsync(string key)` (single argument, no scope) and `ExistsByKeyAsync(string key)` (single argument). On the PR branch these are no longer called from any service method. Leaving dead interface members creates confusion about whether consumers should use them. Either document them as deprecated or remove them and verify no other callers exist.

---

### SUGGESTIONS

#### 7. `ExistsByKeyAndScopeAsync` silently swallows errors and returns `false`

**File:** `XiansAi.Server.Src/Shared/Repositories/SecretVaultRepository.cs` (ExistsByKeyAndScopeAsync catch block)

When the database call throws, the catch block logs and returns `false`, meaning the uniqueness check silently passes and the create/update proceeds. This could create duplicate documents when the database is degraded. Returning `false` (effectively "no duplicate found") from an error handler is dangerous for a uniqueness check. Consider re-throwing or returning a `ServiceResult.InternalServerError` from the service layer instead of silently proceeding.

---

#### 8. `SecretVaultUpdateRequest.Key` uses `null!` default — no XML doc or validation attribute

**Files:** `AdminSecretVaultEndpoints.cs` line ~198, `SecretVaultEndpoints.cs` line ~232

```csharp
public string Key { get; set; } = null!;
```

The `null!` default suppresses the nullable warning but means ASP.NET model binding will bind `null` when the field is absent from the JSON body. The endpoint checks `IsNullOrWhiteSpace(request.Key)` which covers this, but consider adding `[Required]` or using `string Key { get; set; }` initialized to `""` so model validation catches missing keys before handler execution — consistent with how `Key` is declared in the create request.

---

#### 9. No automated tests added for the new key+scope endpoints

The PR adds `.http` scratch files but no automated integration or unit tests for:
- `UpdateByKeyAsync` — success path, key not found, scope mismatch, scope conflict
- `DeleteByKeyAsync` — success path, key not found
- Tenant existence validation returning 404
- Concurrent create with the same key+scope (requires unique index to be meaningful)

The existing `XiansAiWebApplicationFactory` and integration test base are already in place. Tests for the `SecretVault` endpoints should be added.

---

### Files Changed

| File | Assessment |
|------|------------|
| `Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs` | Correct direction; inconsistent tenant guard (Warning 3) |
| `Features/AgentApi/Endpoints/SecretVaultEndpoints.cs` | Correct direction; Delete missing tenant guard (Warning 4) |
| `Shared/Repositories/SecretVaultRepository.cs` | `BuildKeyAndScopeFilter` refactor is clean; missing DB-level index (Critical 1); dead interface members (Warning 6) |
| `Shared/Services/SecretVaultService.cs` | Core logic correct; race condition (Critical 1); ambiguous scope in UpdateByKeyAsync (Warning 5); redundant null pattern (Warning 5) |
| `docs/SECRET_VAULT.md` | Accurate and well-written |
| `docs/SECRET_VAULT_VALIDATION.md` | Accurate |
| `http/*.http` | Correctly updated test files |

---

*Automated review generated by Claude code review agent.*

hasith:
**Review in progress** — Coordinating specialized sub-agents (code, security, test, performance). Full report will be posted shortly.

hasith:
## PR Review: Secret Vault Key + Scope Only & Tenant Existence Validation

**Verdict: REQUEST CHANGES**

The PR introduces a well-considered redesign (removing id-based endpoints, enforcing uniqueness on the full key+scope tuple, adding tenant existence validation) but contains a **critical compilation error** in the repository layer that will prevent the build from succeeding, two **security inconsistencies** in tenant validation logic, and a complete absence of automated tests for all new code paths. These issues must be addressed before merging.

---

## CRITICAL

### 1. `ExistsByKeyAsync` — compilation error (wrong return type)

**File:** `XiansAi.Server.Src/Shared/Repositories/SecretVaultRepository.cs`, line ~83

The diff refactors the body of `ExistsByKeyAsync(string key)` (signature with **no scope parameters**) to call `BuildKeyAndScopeFilter(key, tenantId, agentId, userId, activationName)`, where `tenantId`, `agentId`, `userId`, and `activationName` are **not in scope** for that method. Additionally, the patched body ends with `return await _collection.Find(filter).FirstOrDefaultAsync()`, which returns `Task<SecretVault?>` — but the method declaration is `Task<bool>`.

This means the code as written will not compile. The most likely intent was to simplify `ExistsByKeyAsync` to a raw key-only check (since scope-aware uniqueness is now delegated to `ExistsByKeyAndScopeAsync`), but the body was not updated to match the simplified signature.

**Required fix — body should be:**
```csharp
public async Task<bool> ExistsByKeyAsync(string key)
{
    try
    {
        var filter = Builders<SecretVault>.Filter.Eq(x => x.Key, key);
        var count = await _collection.CountDocumentsAsync(filter);
        return count > 0;
    }
    catch (Exception ex)
    {
        _logger.LogError(ex, "Error checking existence for key {Key}", key);
        return false;
    }
}
```

Note: `SecretVaultService.CreateAsync` calls `_repository.ExistsByKeyAndScopeAsync(...)` in the new code — verify that the old `ExistsByKeyAsync` is no longer called anywhere after this PR, or remove it from the interface entirely.

---

### 2. Old id-based endpoints (`GET /{id}`, `PUT /{id}`, `DELETE /{id}`) still present in file on PR branch

**Files:**
- `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs`, lines 109–179 (current HEAD on main)
- `XiansAi.Server.Src/Features/AgentApi/Endpoints/SecretVaultEndpoints.cs`, lines 110–180 (current HEAD on main)

The diff removes these endpoints in the PR patch, but the PR description says they are removed while the GitHub PR diff shows the expected removals. Confirm that the branch state actually removes all three `GET/{id}`, `PUT/{id}`, and `DELETE/{id}` endpoint registrations from both files and that no routing conflicts exist between the new `MapPut("")` and `MapDelete("")` and the existing `MapPost("")` on the same group.

---

## WARNINGS

### 3. Asymmetric tenant existence check: Admin API vs Agent API (create endpoint)

**Files:**
- `XiansAi.Server.Src/Features/AgentApi/Endpoints/SecretVaultEndpoints.cs` — create (POST)
- `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs` — create (POST)

The **Agent API** create endpoint validates tenant existence whenever `effectiveTenantId` is non-empty:
```csharp
if (!string.IsNullOrWhiteSpace(effectiveTenantId))
{
    var tenant = await tenantCacheService.GetByTenantIdAsync(effectiveTenantId, cancellationToken);
    if (tenant == null) return ... 404 ...
}
```

The **Admin API** create endpoint requires both `request.TenantId` AND `effectiveTenantId` to be non-empty:
```csharp
if (!string.IsNullOrWhiteSpace(request.TenantId) && !string.IsNullOrWhiteSpace(effectiveTenantId))
```

A SysAdmin whose token carries a context tenant can create a tenant-scoped secret without having `request.TenantId` set, bypassing the tenant existence check because `request.TenantId` is null while `effectiveTenantId` is non-empty (from `TryResolveScope`). The same asymmetric condition applies to the Admin API update endpoint. The Agent API and Admin API should enforce the same rule. Fix the Admin API condition to match the Agent API:
```csharp
if (!string.IsNullOrWhiteSpace(effectiveTenantId))
```

The same asymmetric condition applies to the Admin API **update** endpoint (line ~143 of the patched file).

---

### 4. `TenantCacheService.GetByTenantIdAsync` throws `ValidationException` for invalid tenant ID format — unhandled at call sites

**File:** `XiansAi.Server.Src/Shared/Services/TenantCacheService.cs`, line 68

`GetByTenantIdAsync` calls `Tenant.SanitizeAndValidateTenantId(tenantId)` which throws a `ValidationException` if the tenant ID does not match the expected format regex. All four new call sites (Agent create, Agent fetch, Agent update, Admin create, Admin update) do not catch this exception.

An attacker or misconfigured client that supplies a malformed `tenantId` will receive an unhandled 500 rather than a clean 400/404. The `TenantCacheService` should catch `ValidationException` internally and return null (treating an unrecognisable tenant ID as "not found"), or each call site should wrap the call in a try/catch and return a 400 Bad Request.

---

### 5. TOCTOU race in `UpdateByKeyAsync` / `DeleteByKeyAsync`

**File:** `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs`

`UpdateByKeyAsync` does:
1. `FindForAccessAsync(key, scope)` — find the entity
2. `UpdateAsync(entity.Id, ...)` — inside `UpdateAsync`: re-fetch by id, apply changes, `ExistsByKeyAndScopeAsync`, then save

`DeleteByKeyAsync` does:
1. `FindForAccessAsync(key, scope)` — find the entity
2. `DeleteAsync(entity.Id)` — delete by id

Between step 1 and step 2 there is no transaction or optimistic concurrency lock. In a concurrent environment (e.g. two requests deleting the same key simultaneously) both could find the entity in step 1, then step 2 could silently no-op for the second caller (MongoDB delete returns `DeletedCount == 0`, which is mapped to a 404 by `DeleteAsync`). For `UpdateAsync`, the race is more subtle: a concurrent create with the same scope could insert between the two calls. For a secret vault this is a low-risk but real gap — at minimum add a comment explaining the known limitation, and consider using MongoDB's `FindOneAndDelete`/`FindOneAndReplace` for true atomicity.

---

### 6. `SecretVaultUpdateRequest.Key` and `AgentSecretVaultUpdateRequest.Key` lack `[Required]` data annotation

**Files:**
- `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminSecretVaultEndpoints.cs` — `SecretVaultUpdateRequest`
- `XiansAi.Server.Src/Features/AgentApi/Endpoints/SecretVaultEndpoints.cs` — `AgentSecretVaultUpdateRequest`

Both DTOs declare `Key` as `string Key { get; set; } = null!;`. Without `[Required]`, ASP.NET Core model binding will silently bind a missing or null `key` field in the JSON body to `null`, bypassing the `string.IsNullOrWhiteSpace(request.Key)` guard only if the runtime nullable-check path is not reached. Adding `[Required]` and `[JsonRequired]` (System.Text.Json) ensures the framework rejects the request with a 400 before the handler runs.

---

## SUGGESTIONS

### 7. `ExistsByKeyAndScopeAsync` uses `CountDocumentsAsync` — consider `Find(limit:1).AnyAsync()`

**File:** `XiansAi.Server.Src/Shared/Repositories/SecretVaultRepository.cs`

`CountDocumentsAsync` performs a full collection scan on MongoDB unless the filter matches a covering index, and it always counts all matching documents even though only the existence of one is needed. Using `Find(filter).Limit(1).AnyAsync()` (or `CountDocumentsAsync(filter, new CountOptions { Limit = 1 })`) will short-circuit after finding the first match and be measurably faster at scale.

```csharp
// Instead of:
var count = await _collection.CountDocumentsAsync(filter);
return count > 0;

// Prefer:
return await _collection.Find(filter).Limit(1).AnyAsync();
```

---

### 8. `UpdateByKeyAsync` performs 3–4 database round-trips

**File:** `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs`

The call chain `UpdateByKeyAsync → FindForAccessAsync → UpdateAsync → GetByIdAsync → ExistsByKeyAndScopeAsync → UpdateAsync (repo)` results in at minimum 3 and up to 4 separate database calls per update. For a hot path (agent fetching secrets on every invocation), this is notable. Consider refactoring `UpdateAsync(string id, ...)` to accept an already-loaded entity (the one found by `FindForAccessAsync`) to eliminate the redundant `GetByIdAsync` call inside `UpdateAsync`.

---

### 9. Redundant double-null-check pattern in `UpdateByKeyAsync`

**File:** `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs`, lines ~307–310

```csharp
var lookupTenantId = input.TenantId == null ? null : (string.IsNullOrWhiteSpace(input.TenantId) ? null : input.TenantId);
```

`string.IsNullOrWhiteSpace` already returns `true` for `null`, so the outer `== null ? null :` guard is redundant. Simplify to:
```csharp
var lookupTenantId = string.IsNullOrWhiteSpace(input.TenantId) ? null : input.TenantId;
```
Apply the same simplification to `lookupAgentId`, `lookupUserId`, and `lookupActivationName`.

---

### 10. `SECRET_VAULT.md` top-level heading downgraded from `#` to `##`

**File:** `XiansAi.Server.Src/docs/SECRET_VAULT.md`, line 1

The change converts the document's top-level heading `# Secret Vault` to `## Secret Vault`. This is likely unintentional — if this file is rendered standalone (e.g. on a docs site), it should retain an `h1` heading. Revert to `#` unless the document is being embedded inside another page that already provides an `h1`.

---

### 11. No automated tests for new code paths

**Test directory:** `XiansAi.Server.Tests/`

The entire change — `UpdateByKeyAsync`, `DeleteByKeyAsync`, `ExistsByKeyAndScopeAsync`, `BuildKeyAndScopeFilter`, tenant existence validation in create/update/fetch, and the asymmetric Admin vs Agent create guards — has no corresponding unit tests or integration tests. The only coverage is two `.http` files (manual test scripts). Given that the PR is fixing a scope-level uniqueness bug, the absence of tests for the exact conflict scenarios (create duplicate scope, update to collide scope, delete non-existent key) means regressions can reappear silently.

At minimum, the following unit tests should be added for `SecretVaultService`:
- Create: duplicate key+scope returns 409; same key different scope returns 201
- Update: scope collision returns 409; valid update by key+scope succeeds
- Delete: delete by key+scope succeeds; delete non-existent key returns 404
- `UpdateByKeyAsync`: delegates correctly to `UpdateAsync` after lookup; lookup-not-found returns 404
- Tenant existence check: non-existent tenant returns 404; valid tenant allows operation to proceed

---

## Summary by Category

| Category | Finding | Severity |
|---|---|---|
| Code Quality | `ExistsByKeyAsync` body compiles to wrong return type (scope vars not in scope, returns `SecretVault?` not `bool`) | CRITICAL |
| Code Quality | Old id-based endpoints must be confirmed removed from PR branch | CRITICAL |
| Security | Admin API tenant existence check has stricter condition than Agent API — SysAdmin can bypass when `request.TenantId` is null | WARNING |
| Security | `ValidationException` from `TenantCacheService` is unhandled at all new call sites — causes 500 on malformed tenantId | WARNING |
| Security | TOCTOU race in `UpdateByKeyAsync` / `DeleteByKeyAsync` between find and mutate | WARNING |
| Code Quality | `Key` property on update DTOs lacks `[Required]` attribute | WARNING |
| Performance | `CountDocumentsAsync` counts all matching documents; `Find.Limit(1).AnyAsync()` is more efficient | SUGGESTION |
| Performance | `UpdateByKeyAsync` triggers 3-4 DB round-trips due to delegation chain | SUGGESTION |
| Code Quality | Redundant double-null-check pattern in `UpdateByKeyAsync` normalization | SUGGESTION |
| Docs | `SECRET_VAULT.md` heading accidentally downgraded from `#` to `##` | SUGGESTION |
| Testing | Zero automated tests for all new service methods, repository methods, and endpoint logic | WARNING |

PR #388, opened 2026-03-20
OpenTelemetry implementation

prabarshana:
Added open telemetry support for the Xians server

### DIFF
--- a/XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs
+++ b/XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs
@@ -136,6 +136,9 @@ public static class SharedConfiguration
             
         // Add infrastructure services (clients, data access, etc.)
         builder.Services.AddInfrastructureServices(builder.Configuration);
+
+        // Add OpenTelemetry observability (tracing and metrics)
+        builder.AddOpenTelemetry();
         
         // Add common api-related services
         builder.Services.AddEndpointsApiExplorer();
--- a/XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs
+++ b/XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs
@@ -279,11 +279,9 @@ public static class SharedConfiguration
 
         // After auth resolves the tenant, push the tenant tag into the logging scope so every
         // log record emitted within a request carries it as a structured attribute.
-        // Tag name is configurable via OPENTELEMETRY_TENANT_TAG_NAME (default: tenant.id).
-        var tenantTagName =
-            (app.Configuration.GetValue<string>("OpenTelemetry:TenantTagName")
-             ?? Environment.GetEnvironmentVariable("OPENTELEMETRY_TENANT_TAG_NAME"))?.Trim()
-            is { Length: > 0 } t ? t : "tenant.id";
+        var tenantTagName = OpenTelemetryExtensions.ResolveTenantTagName(app.Configuration);
+        var scopeLogger = app.Services.GetRequiredService<ILoggerFactory>()
+            .CreateLogger("XiansAi.TenantScope");
 
         app.Use(async (context, next) =>
         {
@@ -291,9 +289,7 @@ public static class SharedConfiguration
             var tenantId = tenantContext?.TenantId;
             if (!string.IsNullOrWhiteSpace(tenantId))
             {
-                var loggerFactory = context.RequestServices.GetRequiredService<ILoggerFactory>();
-                var logger = loggerFactory.CreateLogger("XiansAi.TenantScope");
-                using (logger.BeginScope(new Dictionary<string, object> { [tenantTagName] = tenantId }))
+                using (scopeLogger.BeginScope(new Dictionary<string, object> { [tenantTagName] = tenantId }))
                 {
                     await next(context);
                 }
--- a/XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs
+++ b/XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs
@@ -277,6 +277,33 @@ public static class SharedConfiguration
         app.UseAuthentication();
         app.UseAuthorization();
 
+        // After auth resolves the tenant, push the tenant tag into the logging scope so every
+        // log record emitted within a request carries it as a structured attribute.
+        // Tag name is configurable via OPENTELEMETRY_TENANT_TAG_NAME (default: tenant.id).
+        var tenantTagName =
+            (app.Configuration.GetValue<string>("OpenTelemetry:TenantTagName")
+             ?? Environment.GetEnvironmentVariable("OPENTELEMETRY_TENANT_TAG_NAME"))?.Trim()
+            is { Length: > 0 } t ? t : "tenant.id";
+
+        app.Use(async (context, next) =>
+        {
+            var tenantContext = context.RequestServices.GetService<ITenantContext>();
+            var tenantId = tenantContext?.TenantId;
+            if (!string.IsNullOrWhiteSpace(tenantId))
+            {
+                var loggerFactory = context.RequestServices.GetRequiredService<ILoggerFactory>();
+                var logger = loggerFactory.CreateLogger("XiansAi.TenantScope");
+                using (logger.BeginScope(new Dictionary<string, object> { [tenantTagName] = tenantId }))
+                {
+                    await next(context);
+                }
+            }
+            else
+            {
+                await next(context);
+            }
+        });
+
         // Map health checks
         app.MapHealthChecks("/health");
 
--- a/XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs
+++ b/XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics;
 using Shared.Repositories;
 using Shared.Services;
 using Microsoft.AspNetCore.Mvc;
@@ -285,10 +286,18 @@ public static class SharedConfiguration
 
         app.Use(async (context, next) =>
         {
-            var tenantContext = context.RequestServices.GetService<ITenantContext>();
-            var tenantId = tenantContext?.TenantId;
+            // Activity.Current has tenant.id set early by EnrichWithHttpRequest (from X-Tenant-Id
+            // header) before auth even runs. Fall back to ITenantContext for API-key paths that
+            // authenticate without a header (e.g. mTLS / agent certificate auth).
+            var tenantId = Activity.Current?.GetTagItem(tenantTagName) as string;
+            if (string.IsNullOrWhiteSpace(tenantId))
+            {
+                tenantId = context.RequestServices.GetService<ITenantContext>()?.TenantId;
+            }
+
             if (!string.IsNullOrWhiteSpace(tenantId))
             {
+                Activity.Current?.SetTag(tenantTagName, tenantId);
                 using (scopeLogger.BeginScope(new Dictionary<string, object> { [tenantTagName] = tenantId }))
                 {
                     await next(context);
--- a/XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs
+++ b/XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs
@@ -278,36 +278,39 @@ public static class SharedConfiguration
         app.UseAuthentication();
         app.UseAuthorization();
 
-        // After auth resolves the tenant, push the tenant tag into the logging scope so every
-        // log record emitted within a request carries it as a structured attribute.
-        var tenantTagName = OpenTelemetryExtensions.ResolveTenantTagName(app.Configuration);
-        var scopeLogger = app.Services.GetRequiredService<ILoggerFactory>()
-            .CreateLogger("XiansAi.TenantScope");
-
-        app.Use(async (context, next) =>
+        if (app.Configuration.GetValue<bool>("OpenTelemetry:Enabled", false))
         {
-            // Activity.Current has tenant.id set early by EnrichWithHttpRequest (from X-Tenant-Id
-            // header) before auth even runs. Fall back to ITenantContext for API-key paths that
-            // authenticate without a header (e.g. mTLS / agent certificate auth).
-            var tenantId = Activity.Current?.GetTagItem(tenantTagName) as string;
-            if (string.IsNullOrWhiteSpace(tenantId))
-            {
-                tenantId = context.RequestServices.GetService<ITenantContext>()?.TenantId;
-            }
+            // After auth resolves the tenant, push the tenant tag into both Activity and logging
+            // scope so spans and OTel-exported log records carry the tenant attribute consistently.
+            var tenantTagName = OpenTelemetryExtensions.ResolveTenantTagName(app.Configuration);
+            var scopeLogger = app.Services.GetRequiredService<ILoggerFactory>()
+                .CreateLogger("XiansAi.TenantScope");
 
-            if (!string.IsNullOrWhiteSpace(tenantId))
+            app.Use(async (context, next) =>
             {
-                Activity.Current?.SetTag(tenantTagName, tenantId);
-                using (scopeLogger.BeginScope(new Dictionary<string, object> { [tenantTagName] = tenantId }))
+                // Activity.Current has tenant.id set early by EnrichWithHttpRequest (from X-Tenant-Id
+                // header) before auth even runs. Fall back to ITenantContext for API-key paths that
+                // authenticate without a header (e.g. mTLS / agent certificate auth).
+                var tenantId = Activity.Current?.GetTagItem(tenantTagName) as string;
+                if (string.IsNullOrWhiteSpace(tenantId))
+                {
+                    tenantId = context.RequestServices.GetService<ITenantContext>()?.TenantId;
+                }
+
+                if (!string.IsNullOrWhiteSpace(tenantId))
+                {
+                    Activity.Current?.SetTag(tenantTagName, tenantId);
+                    using (scopeLogger.BeginScope(new Dictionary<string, object> { [tenantTagName] = tenantId }))
+                    {
+                        await next(context);
+                    }
+                }
+                else
                 {
                     await next(context);
                 }
-            }
-            else
-            {
-                await next(context);
-            }
-        });
+            });
+        }
 
         // Map health checks
         app.MapHealthChecks("/health");
--- a/XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs
+++ b/XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs
@@ -75,6 +75,7 @@ public class WorkflowSignalService : IWorkflowSignalService
     private readonly ILogger<WorkflowSignalService> _logger;
     private readonly ITenantContext _tenantContext;
     private readonly string _tenantTagName;
+    private readonly bool _includeUserIdentity;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="WorkflowSignalService"/> class.
@@ -98,6 +99,7 @@ public class WorkflowSignalService : IWorkflowSignalService
         _tenantContext = tenantContext ?? throw new ArgumentNullException(nameof(tenantContext));
         _tenantTagName = OpenTelemetryExtensions.ResolveTenantTagName(
             configuration ?? throw new ArgumentNullException(nameof(configuration)));
+        _includeUserIdentity = configuration.GetValue<bool>("OpenTelemetry:IncludeUserIdentity", false);
     }
 
     public async Task<IResult> SignalWithStartWorkflow(WorkflowSignalWithStartRequest request)
@@ -117,7 +119,7 @@ public class WorkflowSignalService : IWorkflowSignalService
                 activity?.SetTag(_tenantTagName, _tenantContext.TenantId);
             }
 
-            if (!string.IsNullOrEmpty(_tenantContext.LoggedInUser))
+            if (_includeUserIdentity && !string.IsNullOrEmpty(_tenantContext.LoggedInUser))
             {
                 activity?.SetTag("user.id", _tenantContext.LoggedInUser);
             }
@@ -149,7 +151,7 @@ public class WorkflowSignalService : IWorkflowSignalService
 
             activity?.SetStatus(ActivityStatusCode.Ok);
 
-            _logger.LogInformation("Successfully onvoked workflow type {WorkflowType} with signal {SignalName}", 
+            _logger.LogInformation("Successfully invoked workflow type {WorkflowType} with signal {SignalName}", 
                 request.TargetWorkflowType, request.SignalName);
             
             return Results.Ok(new { 
--- a/XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs
+++ b/XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs
@@ -4,6 +4,7 @@ using Shared.Auth;
 using Shared.Utils;
 using System.Text.Json.Serialization;
 using Shared.Utils.Temporal;
+using Features.Shared.Configuration;
 using Features.WebApi.Services;
 
 namespace Shared.Services;
@@ -73,6 +74,7 @@ public class WorkflowSignalService : IWorkflowSignalService
     private readonly ITemporalClientFactory _clientFactory;
     private readonly ILogger<WorkflowSignalService> _logger;
     private readonly ITenantContext _tenantContext;
+    private readonly string _tenantTagName;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="WorkflowSignalService"/> class.
@@ -81,17 +83,21 @@ public class WorkflowSignalService : IWorkflowSignalService
     /// <param name="logger">The logger for recording operational information.</param>
     /// <param name="tenantContext">The tenant context for the current request.</param>
     /// <param name="agentService">The agent service for checking if an agent is system scoped.</param>
+    /// <param name="configuration">The application configuration.</param>
     /// <exception cref="ArgumentNullException">Thrown when any of the required services is null.</exception>
     public WorkflowSignalService(
         IAgentService agentService,
         ITemporalClientFactory clientFactory,
         ILogger<WorkflowSignalService> logger,
-        ITenantContext tenantContext)
+        ITenantContext tenantContext,
+        IConfiguration configuration)
     {
         _agentService = agentService ?? throw new ArgumentNullException(nameof(agentService));
         _clientFactory = clientFactory ?? throw new ArgumentNullException(nameof(clientFactory));
         _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         _tenantContext = tenantContext ?? throw new ArgumentNullException(nameof(tenantContext));
+        _tenantTagName = OpenTelemetryExtensions.ResolveTenantTagName(
+            configuration ?? throw new ArgumentNullException(nameof(configuration)));
     }
 
     public async Task<IResult> SignalWithStartWorkflow(WorkflowSignalWithStartRequest request)
@@ -108,7 +114,7 @@ public class WorkflowSignalService : IWorkflowSignalService
 
             if (!string.IsNullOrEmpty(_tenantContext.TenantId))
             {
-                activity?.SetTag("tenant.id", _tenantContext.TenantId);
+                activity?.SetTag(_tenantTagName, _tenantContext.TenantId);
             }
 
             if (!string.IsNullOrEmpty(_tenantContext.LoggedInUser))
@@ -120,7 +126,7 @@ public class WorkflowSignalService : IWorkflowSignalService
 
             activity?.SetTag("temporal.namespace", client.Options.Namespace);
 
-            var systemScoped = _agentService.IsSystemAgent(request.SourceAgent).Result.Data;
+            var systemScoped = (await _agentService.IsSystemAgent(request.SourceAgent)).Data;
 
             var options = new NewWorkflowOptions(
                 request.SourceAgent, 
--- a/XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs
+++ b/XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs
@@ -1,3 +1,5 @@
+using System.Diagnostics;
+using OpenTelemetry.Trace;
 using Shared.Auth;
 using Shared.Utils;
 using System.Text.Json.Serialization;
@@ -65,6 +67,8 @@ public interface IWorkflowSignalService
 
 public class WorkflowSignalService : IWorkflowSignalService
 {
+    private static readonly ActivitySource ActivitySource = new("XiansAi.Server.Temporal");
+
     private readonly IAgentService _agentService;
     private readonly ITemporalClientFactory _clientFactory;
     private readonly ILogger<WorkflowSignalService> _logger;
@@ -92,9 +96,29 @@ public class WorkflowSignalService : IWorkflowSignalService
 
     public async Task<IResult> SignalWithStartWorkflow(WorkflowSignalWithStartRequest request)
     {
+        using var activity = ActivitySource.StartActivity("Temporal.SignalWithStart", ActivityKind.Client);
+
         try
         {
+            activity?.SetTag("temporal.operation", "signal_with_start");
+            activity?.SetTag("temporal.workflow_type", request.TargetWorkflowType);
+            activity?.SetTag("temporal.signal_name", request.SignalName ?? "unknown");
+            activity?.SetTag("temporal.source_agent", request.SourceAgent);
+            activity?.SetTag("temporal.workflow_id", request.TargetWorkflowId ?? "auto-generated");
+
+            if (!string.IsNullOrEmpty(_tenantContext.TenantId))
+            {
+                activity?.SetTag("tenant.id", _tenantContext.TenantId);
+            }
+
+            if (!string.IsNullOrEmpty(_tenantContext.LoggedInUser))
+            {
+                activity?.SetTag("user.id", _tenantContext.LoggedInUser);
+            }
+
             var client = await _clientFactory.GetClientAsync() ?? throw new Exception("Failed to get Temporal client");
+
+            activity?.SetTag("temporal.namespace", client.Options.Namespace);
 
             var systemScoped = _agentService.IsSystemAgent(request.SourceAgent).Result.Data;
 
@@ -117,6 +141,8 @@ public class WorkflowSignalService : IWorkflowSignalService
 
             await client.StartWorkflowAsync(request.TargetWorkflowType, new List<object>().AsReadOnly(), options);
 
+            activity?.SetStatus(ActivityStatusCode.Ok);
+
             _logger.LogInformation("Successfully onvoked workflow type {WorkflowType} with signal {SignalName}", 
                 request.TargetWorkflowType, request.SignalName);
             
@@ -128,6 +154,8 @@ public class WorkflowSignalService : IWorkflowSignalService
         }
         catch (Temporalio.Exceptions.RpcException ex) when (ex.Message.Contains("workflow not found"))
         {
+            activity?.SetStatus(ActivityStatusCode.Error, ex.Message);
+            activity?.RecordException(ex);
             _logger.LogWarning(ex, "Workflow reference not found for type: {WorkflowType}", request.TargetWorkflowType);
             return Results.NotFound(new {
                 message = $"Workflow type '{request.TargetWorkflowType}' could not be started or referenced",
@@ -136,6 +164,8 @@ public class WorkflowSignalService : IWorkflowSignalService
         }
         catch (Exception ex)
         {
+            activity?.SetStatus(ActivityStatusCode.Error, ex.Message);
+            activity?.RecordException(ex);
             _logger.LogError(ex, "Error sending signal {SignalName} to workflow {WorkflowType}", 
                 request.SignalName, request.TargetWorkflowType);
                 
--- a/XiansAi.Server.Src/.env.example
+++ b/XiansAi.Server.Src/.env.example
@@ -131,7 +131,11 @@ OpenTelemetry__OtlpEndpoint=http://aspire-dashboard:18889
 #OPENTELEMETRY_TENANT_TAG_NAME=tenant.id
 # Comma-separated list of MongoDB commands to suppress from traces.
 # When not set, all commands are monitored. Uncomment and adjust as needed.
+#OpenTelemetry__MongoDB__ExcludedCommands=getMore,ping
+# Legacy key (supported for backward compatibility):
 #OPENTELEMETRY_MONGODB_EXCLUDED_COMMANDS=getMore,ping
 # Capture raw MongoDB query documents as span attributes (filter values, $match fields, etc.).
 # OFF by default — query documents can contain tenant data or PII. Enable only in trusted environments.
+#OpenTelemetry__MongoDB__CaptureCommandText=false
+# Legacy key (supported for backward compatibility):
 #OPENTELEMETRY_MONGODB_CAPTURE_COMMAND_TEXT=false
--- a/XiansAi.Server.Src/.env.example
+++ b/XiansAi.Server.Src/.env.example
@@ -113,3 +113,16 @@ Logging__LogLevel__Microsoft.AspNetCore=Warning
 # USAGE EVENTS 
 # ====================
 UsageEvents__RecordUsageEvents=true
+
+# ====================
+# OPENTELEMETRY (Optional)
+# ====================
+# Set Enabled=true and provide an OtlpEndpoint to activate telemetry export.
+# Development (Aspire Dashboard): http://aspire-dashboard:18889
+# Production (OTEL Collector):    http://otel-collector:4317
+OpenTelemetry__Enabled=false
+OpenTelemetry__ServiceName=XiansAi.Server
+OpenTelemetry__OtlpEndpoint=http://aspire-dashboard:18889
+# Comma-separated list of MongoDB commands to suppress from traces.
+# When not set, all commands are monitored. Example: OPENTELEMETRY_MONGODB_EXCLUDED_COMMANDS=getMore,ping
+OPENTELEMETRY_MONGODB_EXCLUDED_COMMANDS=getMore,ping
--- a/XiansAi.Server.Src/.env.example
+++ b/XiansAi.Server.Src/.env.example
@@ -123,9 +123,15 @@ UsageEvents__RecordUsageEvents=true
 OpenTelemetry__Enabled=false
 OpenTelemetry__ServiceName=XiansAi.Server
 OpenTelemetry__OtlpEndpoint=http://aspire-dashboard:18889
+# Emit user identity fields (tenant.user, tenant.user_type, tenant.user_roles) as span attributes.
+# OFF by default — these are PII. Enable only in trusted, internal-only observability environments.
+#OpenTelemetry__IncludeUserIdentity=false
 # Tag name used for tenant identity on all spans and log records (default: tenant.id).
 # Override when your observability backend expects a different convention (e.g. custom.tenant_id).
 #OPENTELEMETRY_TENANT_TAG_NAME=tenant.id
 # Comma-separated list of MongoDB commands to suppress from traces.
-# When not set, all commands are monitored. Example: OPENTELEMETRY_MONGODB_EXCLUDED_COMMANDS=getMore,ping
-OPENTELEMETRY_MONGODB_EXCLUDED_COMMANDS=getMore,ping
+# When not set, all commands are monitored. Uncomment and adjust as needed.
+#OPENTELEMETRY_MONGODB_EXCLUDED_COMMANDS=getMore,ping
+# Capture raw MongoDB query documents as span attributes (filter values, $match fields, etc.).
+# OFF by default — query documents can contain tenant data or PII. Enable only in trusted environments.
+#OPENTELEMETRY_MONGODB_CAPTURE_COMMAND_TEXT=false
--- a/XiansAi.Server.Src/.env.example
+++ b/XiansAi.Server.Src/.env.example
@@ -123,6 +123,9 @@ UsageEvents__RecordUsageEvents=true
 OpenTelemetry__Enabled=false
 OpenTelemetry__ServiceName=XiansAi.Server
 OpenTelemetry__OtlpEndpoint=http://aspire-dashboard:18889
+# Tag name used for tenant identity on all spans and log records (default: tenant.id).
+# Override when your observability backend expects a different convention (e.g. custom.tenant_id).
+#OPENTELEMETRY_TENANT_TAG_NAME=tenant.id
 # Comma-separated list of MongoDB commands to suppress from traces.
 # When not set, all commands are monitored. Example: OPENTELEMETRY_MONGODB_EXCLUDED_COMMANDS=getMore,ping
 OPENTELEMETRY_MONGODB_EXCLUDED_COMMANDS=getMore,ping
--- a/XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs
+++ b/XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs
@@ -1,0 +1,155 @@
+using OpenTelemetry.Logs;
+using OpenTelemetry.Metrics;
+using OpenTelemetry.Resources;
+using OpenTelemetry.Trace;
+using Shared.Auth;
+
+namespace Features.Shared.Configuration;
+
+/// <summary>
+/// Configures OpenTelemetry tracing and metrics for XiansAi.Server.
+/// Enabled via OpenTelemetry:Enabled=true; exports via OTLP gRPC to OpenTelemetry:OtlpEndpoint.
+/// </summary>
+public static class OpenTelemetryExtensions
+{
+    public static WebApplicationBuilder AddOpenTelemetry(this WebApplicationBuilder builder)
+    {
+        var enabled = builder.Configuration.GetValue<bool>("OpenTelemetry:Enabled", false);
+        if (!enabled)
+        {
+            return builder;
+        }
+
+        var serviceName = builder.Configuration.GetValue<string>("OpenTelemetry:ServiceName") ?? "XiansAi.Server";
+        var serviceVersion = typeof(Program).Assembly.GetName().Version?.ToString() ?? "1.0.0";
+        var otlpEndpoint = builder.Configuration.GetValue<string>("OpenTelemetry:OtlpEndpoint");
+
+        if (string.IsNullOrWhiteSpace(otlpEndpoint))
+        {
+            Console.WriteLine("[OpenTelemetry] WARNING: OpenTelemetry:OtlpEndpoint is not set — telemetry export disabled.");
+            return builder;
+        }
+
+        Console.WriteLine($"[OpenTelemetry] Initializing for service: {serviceName}");
+        Console.WriteLine($"[OpenTelemetry] OTLP Endpoint: {otlpEndpoint}");
+
+        try
+        {
+            builder.Services.AddOpenTelemetry()
+                .ConfigureResource(resource => resource
+                    .AddService(
+                        serviceName: serviceName,
+                        serviceVersion: serviceVersion,
+                        serviceInstanceId: Environment.MachineName)
+                    .AddAttributes(new Dictionary<string, object>
+                    {
+                        ["deployment.environment"] = builder.Environment.EnvironmentName,
+                        ["host.name"] = Environment.MachineName,
+                    }))
+                .WithTracing(tracing => tracing
+                    .AddAspNetCoreInstrumentation(options =>
+                    {
+                        options.RecordException = true;
+
+                        // Exclude health check endpoints to reduce noise
+                        options.Filter = context =>
+                            !context.Request.Path.StartsWithSegments("/health");
+
+                        // Tag incoming requests with tenant context from HTTP headers
+                        options.EnrichWithHttpRequest = (activity, httpRequest) =>
+                        {
+                            activity.SetTag("http.method", httpRequest.Method);
+                            activity.SetTag("http.scheme", httpRequest.Scheme);
+                            activity.SetTag("http.host", httpRequest.Host.Value);
+
+                            var tenantId = httpRequest.Headers["X-Tenant-Id"].FirstOrDefault();
+                            if (!string.IsNullOrEmpty(tenantId))
+                            {
+                                activity.SetTag("tenant.id", tenantId);
+                            }
+                        };
+
+                        // Enrich with authenticated tenant context after auth middleware runs
+                        options.EnrichWithHttpResponse = (activity, httpResponse) =>
+                        {
+                            activity.SetTag("http.status_code", httpResponse.StatusCode);
+
+                            var tenantContext = httpResponse.HttpContext.RequestServices.GetService<ITenantContext>();
+                            if (tenantContext != null && !string.IsNullOrEmpty(tenantContext.TenantId))
+                            {
+                                activity.SetTag("tenant.id", tenantContext.TenantId);
+                                activity.SetTag("tenant.user", tenantContext.LoggedInUser);
+                                activity.SetTag("tenant.user_type", tenantContext.UserType.ToString());
+
+                                if (tenantContext.UserRoles?.Length > 0)
+                                {
+                                    activity.SetTag("tenant.user_roles", string.Join(",", tenantContext.UserRoles));
+                                }
+                            }
+                        };
+                    })
+                    .AddHttpClientInstrumentation(options =>
+                    {
+                        options.RecordException = true;
+
+                        options.EnrichWithHttpRequestMessage = (activity, request) =>
+                        {
+                            activity.SetTag("http.request.method", request.Method.ToString());
+
+                            var tenantId = request.Headers.TryGetValues("X-Tenant-Id", out var values)
+                                ? values.FirstOrDefault()
+                                : null;
+                            if (!string.IsNullOrEmpty(tenantId))
+                            {
+                                activity.SetTag("tenant.id", tenantId);
+                            }
+                        };
+
+                        options.EnrichWithHttpResponseMessage = (activity, response) =>
+                            activity.SetTag("http.response.status_code", (int)response.StatusCode);
+                    })
+                    .AddSource("XiansAi.*")
+                    .AddSource("MongoDB.Driver.Core.Extensions.DiagnosticSources")
+                    .AddOtlpExporter(options =>
+                    {
+                        options.Endpoint = new Uri(otlpEndpoint);
+                        options.Protocol = OpenTelemetry.Exporter.OtlpExportProtocol.Grpc;
+                    }))
+                .WithMetrics(metrics => metrics
+                    .AddAspNetCoreInstrumentation()
+                    .AddHttpClientInstrumentation()
+                    .AddRuntimeInstrumentation()
+                    .AddMeter("XiansAi.*")
+                    .AddOtlpExporter(options =>
+                    {
+                        options.Endpoint = new Uri(otlpEndpoint);
+                        options.Protocol = OpenTelemetry.Exporter.OtlpExportProtocol.Grpc;
+                    }));
+
+            // Logs use builder.Logging.AddOpenTelemetry() because IncludeFormattedMessage / ParseStateValues /
+            // IncludeScopes are on OpenTelemetryLoggerOptions, not on LoggerProviderBuilder (WithLogging parameter).
+            builder.Logging.AddOpenTelemetry(logging =>
+            {
+                logging.IncludeFormattedMessage = true;
+                logging.ParseStateValues = true;
+                logging.IncludeScopes = true;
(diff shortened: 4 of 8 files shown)


xianix-agent (changes requested):
# PR Review Report

**PR:** Add OpenTelemetry support for tracing and metrics / Enhance OpenTelemetry configuration to support customizable tenant tagging
**Author:** Isanka Patabandige &lt;isankap@99x.io&gt;
**Files Changed:** 8 | +330 / -3

---

## Summary

This PR adds a well-structured OpenTelemetry observability layer covering tracing, metrics, and structured logging across ASP.NET Core, MongoDB, and Temporal. The feature is safely opt-in (disabled by default), the catch-all around OTel startup prevents telemetry failures from crashing the server, and tenant context propagation is correctly applied at both the HTTP and MongoDB span levels. However, there are several blocking issues around PII exposure via span tags, a blocking `.Result` call that can cause deadlocks in production, a `CaptureCommandText = true` flag that sends raw MongoDB command text to the collector unconditionally, and a significant gap in test coverage that must be addressed before merge.

---

## Critical Issues (Must Fix)

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:123` — **Sync-over-async deadlock risk.** `_agentService.IsSystemAgent(...).Result` blocks the async thread and can cause deadlocks under ASP.NET Core's synchronization context. The surrounding method is already `async`; this should be awaited.

  ```csharp
  // Current (problematic)
  var systemScoped = _agentService.IsSystemAgent(request.SourceAgent).Result.Data;

  // Fix
  var systemScoped = (await _agentService.IsSystemAgent(request.SourceAgent)).Data;
  ```

- [ ] `XiansAi.Server.Src/Shared/Data/MongoDbClientService.cs:87` — **`CaptureCommandText = true` sends raw MongoDB queries (including filter values) to the OTLP collector.** This can leak tenant data, PII, or secrets (e.g., plaintext field values in `$match` or `findOne` filters) to any observer of the telemetry backend. This should be opt-in and explicitly documented.

  ```csharp
  // Current (problematic)
  CaptureCommandText = true,

  // Fix — read from configuration so operators can opt in consciously
  CaptureCommandText = Environment.GetEnvironmentVariable("OPENTELEMETRY_MONGODB_CAPTURE_COMMAND_TEXT")
      ?.Equals("true", StringComparison.OrdinalIgnoreCase) ?? false,
  ```

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:89-90` — **`tenant.user` and `tenant.user_roles` tags on HTTP response spans expose PII to the OTLP collector.** Authenticated user identity and role lists are sensitive. These should either be omitted entirely, placed behind a configuration flag, or hashed.

  ```csharp
  // Current (problematic)
  activity.SetTag("tenant.user", tenantContext.LoggedInUser);
  activity.SetTag("tenant.user_type", tenantContext.UserType.ToString());
  // ...
  activity.SetTag("tenant.user_roles", string.Join(",", tenantContext.UserRoles));

  // Fix — guard with a configuration flag, e.g. OpenTelemetry:IncludeUserIdentity
  if (includePiiTags)
  {
      activity.SetTag("tenant.user", tenantContext.LoggedInUser);
      // ...
  }
  ```

---

## Warnings (Should Fix)

- [x] `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs:294-295` — **A new `ILoggerFactory` and `ILogger` are created on every request inside the tenant scope middleware.** Logger instances are cheap once resolved but `CreateLogger` still allocates a new logger object per call. Inject `ILogger<T>` or `ILoggerFactory` as a singleton outside the `app.Use` lambda.

  ```csharp
  // Current (problematic) — inside app.Use lambda, per-request
  var loggerFactory = context.RequestServices.GetRequiredService<ILoggerFactory>();
  var logger = loggerFactory.CreateLogger("XiansAi.TenantScope");

  // Fix — resolve once before the lambda
  var scopeLogger = app.Services.GetRequiredService<ILoggerFactory>()
      .CreateLogger("XiansAi.TenantScope");
  app.Use(async (context, next) =>
  {
      // use scopeLogger
  });
  ```

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:30-33` and `SharedConfiguration.cs:283-286` — **The tenant tag name resolution logic is duplicated across two files.** Both `AddOpenTelemetry` and `UseSharedMiddleware` independently parse `OpenTelemetry:TenantTagName` / `OPENTELEMETRY_TENANT_TAG_NAME`. A shared helper or a named options class should own this logic to prevent drift.

- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/NewWorkflowOptions.cs:113-123` — **`AddToMemo` silently no-ops when `Memo` is `IReadOnlyDictionary`.** The comment says it is "used by the TracingInterceptor", but if `Memo` was ever set to a read-only dictionary (e.g., from an external source), the `else` branch creates a new dict and assigns to `Memo`, which replaces the whole memo. The `else` branch scenario should be documented explicitly or made impossible by enforcing `Dictionary<string, object>` in the constructor.

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:41-42` and `152-153` — **Startup telemetry uses `Console.WriteLine` instead of `ILogger`.** The rest of the codebase uses structured logging. Using `Console.WriteLine` bypasses the logging pipeline, preventing these messages from appearing in any log aggregation backend.

- [ ] `XiansAi.Server.Src/.env.example:131` — **`OPENTELEMETRY_MONGODB_EXCLUDED_COMMANDS=getMore,ping` is not commented out in the example file.** Unlike `OPENTELEMETRY_TENANT_TAG_NAME` (which is commented), this value is set as a default. If copied verbatim into a real `.env` file this silently suppresses `getMore` and `ping` monitoring. It should either be commented out or documented prominently as an active default.

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:111` — **The tenant tag `tenant.id` is hardcoded in `WorkflowSignalService` while the rest of the codebase uses the configurable `tenantTagName`.** This means if an operator overrides the tag name, the Temporal span will still emit `tenant.id` instead of the custom tag, breaking consistency.

  ```csharp
  // Current (problematic)
  activity?.SetTag("tenant.id", _tenantContext.TenantId);

  // Fix — inject the configured tag name or read from a shared constant
  activity?.SetTag(tenantTagName, _tenantContext.TenantId);
  ```

---

## Suggestions (Consider Improving)

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:57-136` — Consider splitting `WithTracing`, `WithMetrics`, and logging setup into separate private methods to reduce the method length of `AddOpenTelemetry` (~130 lines). This improves testability and readability.

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:171-201` — `MongoDbTenantPropagationProcessor` walks the entire parent activity chain on every MongoDB `OnStart`. For deeply nested call stacks this is O(depth). Consider caching the resolved tenant ID in `AsyncLocal<string?>` or limiting the depth of traversal.

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:146` — Typo in log message: `"Successfully onvoked workflow type"` should be `"Successfully invoked workflow type"`. (Pre-existing, but touching this line is a good opportunity to fix it.)

- [ ] `XiansAi.Server.Src/XiansAi.Server.csproj` — The BOM character was removed from the first line (`﻿` → `<Project`). This is harmless but worth confirming it was intentional and does not affect any tooling that expects it.

- [ ] Consider adding a `/metrics` or `/telemetry` health endpoint that returns whether OTel is enabled and what the configured endpoint is. This aids operators diagnosing misconfiguration without needing to restart the service.

---

## Review Details

### Code Quality

The overall structure is clean and follows established patterns in the codebase. The opt-in guard at the top of `AddOpenTelemetry` and the `try/catch` safety wrapper are good defensive practices. The `MongoDbTenantPropagationProcessor` is well-documented. The main structural concerns are: the duplicated tag-name resolution logic (see Warnings), overlong method body in `AddOpenTelemetry`, and the hardcoded `"tenant.id"` string in `WorkflowSignalService` that bypasses the configurable tag system. The `AddToMemo` method in `NewWorkflowOptions` is functional but relies on an implicit type assumption (`Memo is Dictionary<string, object>`) that should be made explicit.

### Security

Two issues require attention before merge. First, `CaptureCommandText = true` in `MongoDbClientService` sends full MongoDB command documents — including filter values and field contents — to the OTLP backend. In a multi-tenant SaaS context this can include tenant data or PII from queries. This flag should be opt-in. Second, `tenant.user` and `tenant.user_roles` are emitted as span tags visible to anyone with access to the observability backend. These should be behind a configuration gate (`OpenTelemetry:IncludePiiTags=false` by default) consistent with least-privilege telemetry practices.

### Test Coverage

No unit or integration tests were added for any of the new code paths:
- `OpenTelemetryExtensions.AddOpenTelemetry` (feature flag guard, endpoint validation, tag resolution)
- `MongoDbTenantPropagationProcessor.OnStart` (parent chain traversal)
- `WorkflowSignalService` tracing paths (activity creation, tag setting, status codes)
- `NewWorkflowOptions.AddToMemo` (both mutable and immutable dictionary paths)

At a minimum, unit tests should cover: (1) that OTel is not registered when `Enabled=false`, (2) that the tenant tag name falls back to `tenant.id` when neither config source is set, (3) that `MongoDbTenantPropagationProcessor` correctly propagates from parent spans, and (4) that `AddToMemo` handles both the mutable and copy-and-replace paths.

### Performance

The `MongoDbTenantPropagationProcessor` traverses the parent activity chain on every MongoDB span start. For typical request depths (2–5 levels) this is negligible, but it could become a concern in highly nested call stacks. The per-request logger creation in the tenant scope middleware allocates on every request and should be pre-resolved. The triple OTLP exporter configuration (traces, metrics, logs all configured separately) is correct and idiomatic for the current OTel SDK version. The `TracingInterceptor` addition to Temporal is correctly documented as a no-op when no `TracerProvider` is active, so it is safe to leave always-enabled.

---

## Files Reviewed

| `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs` | +203 | High | New OTel setup, PII tag exposure, no tests |
| `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs` | +30 | Medium | Per-request logger allocation, duplicated tag resolution |
| `XiansAi.Server.Src/Shared/Data/MongoDbClientService.cs` | +19 | High | `CaptureCommandText=true` leaks query content |
| `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs` | +30 | High | Sync-over-async `.Result`, hardcoded tag name |
| `XiansAi.Server.Src/Shared/Utils/Temporal/NewWorkflowOptions.cs` | +19/-3 | Low | `AddToMemo` logic reasonable, silent no-op risk |
| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs` | +4 | Low | TracingInterceptor addition is clean and safe |
| `XiansAi.Server.Src/XiansAi.Server.csproj` | +12/-1 | Low | BOM removal and new package refs look correct |
| `XiansAi.Server.Src/.env.example` | +16 | Low | Active default for excluded commands should be commented |

xianix-agent (changes requested):
## PR Review Report

**PR:** Add OpenTelemetry support for tracing and metrics (branch: `pr-388-675`)
**Author:** Isanka Patabandige <isankap@99x.io>
**Files Changed:** 8 | **+368** / **-8**

---

### Summary

This PR introduces a well-structured OpenTelemetry integration covering OTLP tracing, metrics, and logging export for the XiansAi.Server. The implementation is thoughtfully designed with correct PII guards, a configurable tenant tag, and resilient startup behaviour (OTel failures never crash the server). There are a few correctness and design issues that should be resolved before merge, primarily around a logging-scope misuse that silently drops tenant context and a `user.id` tag that bypasses the PII guard.

---

- [ ] `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs:286-301` — **Tenant logging scope is attached to the wrong logger and will not propagate to request-handling code.**

  `scopeLogger.BeginScope(...)` opens a scope on the dedicated `XiansAi.TenantScope` logger. `ILogger.BeginScope` affects only the logger instance it is called on, not the ambient logging pipeline. Log records written by other loggers (`ILogger<MyService>`, etc.) inside `next(context)` will not carry the tenant attribute. The scope must be opened through `ILoggerFactory` in a way that is visible to the logging infrastructure, or — simpler — pushed via `Activity.Current` tagging or an `IEnrichDiagnosticContext` mechanism.

  ```csharp
  // Current (broken) — scope is on a private logger, invisible to other loggers
  using (scopeLogger.BeginScope(new Dictionary<string, object> { [tenantTagName] = tenantId }))
  {
      await next(context);
  }

  // Fix — use HttpContext.Items to push the value into an enricher, or use
  // a custom ILoggerProvider / LoggingMiddleware that reads from ITenantContext.
  // Alternatively, set a tag on the current Activity so all OTel log records pick it up:
  var current = Activity.Current;
  current?.SetTag(tenantTagName, tenantId);
  await next(context);
  ```

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:120-123` — **`user.id` tag is written unconditionally, bypassing the `IncludeUserIdentity` PII guard.**

  The `includeUserIdentity` flag deliberately gates user identity fields in the HTTP instrumentation enrichment. The Temporal span sets `user.id` from `_tenantContext.LoggedInUser` without any such guard, creating an inconsistency: the same data is protected in one path and exposed in another.

  ```csharp
  // Current (PII leak — no guard)
  if (!string.IsNullOrEmpty(_tenantContext.LoggedInUser))
  {
      activity?.SetTag("user.id", _tenantContext.LoggedInUser);
  }

  // Fix — inject the IncludeUserIdentity flag via IConfiguration (same pattern
  // already used for _tenantTagName) and gate the tag behind it
  if (_includeUserIdentity && !string.IsNullOrEmpty(_tenantContext.LoggedInUser))
  {
      activity?.SetTag("user.id", _tenantContext.LoggedInUser);
  }
  ```

---

### Warnings (Should Fix)

- [ ] `XiansAi.Server.Src/Shared/Data/MongoDbClientService.cs:77-88` — **MongoDB instrumentation config reads environment variables directly, inconsistent with the rest of the codebase.**

  Every other configuration value in this service is read from `IConfiguration` (which unifies env vars, appsettings, secrets, etc.). Reading `OPENTELEMETRY_MONGODB_EXCLUDED_COMMANDS` and `OPENTELEMETRY_MONGODB_CAPTURE_COMMAND_TEXT` directly via `Environment.GetEnvironmentVariable` bypasses the standard configuration pipeline, makes the values untestable via `appsettings.json`, and is inconsistent with the `.env.example` pattern used elsewhere for configuration keys. These should be injected via `IConfiguration` under the `OpenTelemetry:` section, matching the pattern used in `OpenTelemetryExtensions.cs`. Note also that `MongoDbClientService` does not currently receive `IConfiguration`; passing it in or moving the OTel option resolution upstream (e.g., into `OpenTelemetryExtensions.AddOpenTelemetry`) would be the right fix.

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:47-49` — **Startup bootstrap logger is created with `LogLevel.Debug` which will write verbose debug output unconditionally during server startup.**

  The intent is clearly a short-lived bootstrap logger, but `SetMinimumLevel(LogLevel.Debug)` means any library that uses this factory will emit debug-level records to the console regardless of the host's configured log levels. Use `LogLevel.Information` (or read it from the host configuration) so startup output respects the operator's intent.

  ```csharp
  // Current
  lb.SetMinimumLevel(LogLevel.Debug).AddSimpleConsole()

  // Fix
  lb.SetMinimumLevel(LogLevel.Information).AddSimpleConsole()
  ```

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:152` — **Typo in log message: "onvoked" should be "invoked".**

  ```csharp
  // Current
  _logger.LogInformation("Successfully onvoked workflow type {WorkflowType} with signal {SignalName}", ...)

  // Fix
  _logger.LogInformation("Successfully invoked workflow type {WorkflowType} with signal {SignalName}", ...)
  ```

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:22-25` — **`ResolveTenantTagName` is a `public static` method on the extensions class, creating a cross-cutting coupling.**

  `WorkflowSignalService` and `SharedConfiguration` both call `OpenTelemetryExtensions.ResolveTenantTagName(configuration)` directly. This creates a hard dependency on the OTel extensions class from unrelated service classes, and means the resolution logic is not injectable or testable in isolation. Consider extracting this into a registered singleton option (e.g., `TelemetryOptions`) that is resolved once at startup and injected where needed via `IOptions<TelemetryOptions>`.

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:88-91` — **`http.method`, `http.scheme`, and `http.host` tags duplicate attributes that ASP.NET Core OTel instrumentation already sets by default.**

  OpenTelemetry's ASP.NET Core instrumentation automatically sets semantic convention attributes including `http.request.method` (OTel 1.7+ semconv) and `server.address`. Manually re-setting `http.method`, `http.scheme`, and `http.host` using the old attribute names (pre-semconv) creates duplicate or conflicting attributes on spans. Remove the redundant tags from `EnrichWithHttpRequest` and rely on the instrumentation's defaults.

---

- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/NewWorkflowOptions.cs:12,54-55` — The `_memo` backing field and `AddToMemo` public method are added in this PR but `AddToMemo` is not called anywhere in the PR diff. The comment says "Used by the TracingInterceptor to inject trace context," but the `TracingInterceptor` from `Temporalio.Extensions.OpenTelemetry` handles context propagation automatically via the Temporal SDK's header mechanism — it does not need to mutate the memo. Consider verifying whether `AddToMemo` is actually needed; if not, it adds unused public surface area.

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:158-162` — Consider adding a `SetMaxExportBatchSize` / `SetScheduledDelayMilliseconds` override or at least a comment noting that OTLP export uses the SDK defaults, so operators know what to tune for high-throughput environments.

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:192-229` — `MongoDbTenantPropagationProcessor` walks the full parent `Activity` chain on every MongoDB span `OnStart`. In high-throughput scenarios with deep call stacks this could add latency to every DB call. The span enrichment in `EnrichWithHttpResponse` already sets the tag on the request span. Consider caching the resolved tenant ID in `Activity.Baggage` or a well-known custom property so the walk is O(1).

- [ ] `XiansAi.Server.Src/XiansAi.Server.csproj:50-56` — All OpenTelemetry packages are pinned to `1.10.0`. Add a comment or a `Directory.Packages.props` entry noting the OTel SDK version in use, so future upgrades are coordinated across all packages (the SDK, exporters, and instrumentation libraries must be version-aligned).

---

### Review Details

#### Code Quality

The overall structure is clean and follows existing patterns well. The use of a `try/catch` around the entire OTel initialization to prevent server crashes is a correct and important safety net. The `MongoDbTenantPropagationProcessor` is a good solution for the tag inheritance gap, though the performance note above applies. The `AddToMemo` public method appears to be dead code in this PR and should be audited. One pre-existing typo ("onvoked") was introduced or carried over in the log message at line 152 of `WorkflowSignalService.cs`.

#### Security

Two findings of note. First, `user.id` on Temporal spans (WorkflowSignalService line 120-123) emits `LoggedInUser` without any PII gate, while the HTTP enrichment path correctly requires `IncludeUserIdentity=true`. This is an inconsistency that could expose user identity to the observability backend unexpectedly. Second, the overall PII design (opt-in identity fields, opt-in command text, default-off) is sound and correctly documented. The `X-Tenant-Id` header is read directly and placed into span tags at line 93-97 — if this header can be spoofed by external callers before auth runs, untrusted data could appear in traces; the response enrichment (post-auth) correctly overwrites with the authenticated value, so this is an acceptable defence-in-depth concern rather than a critical bug.

#### Test Coverage

No new tests were added for any of the introduced functionality. The `OpenTelemetryExtensions`, `MongoDbTenantPropagationProcessor`, and `WorkflowSignalService` activity instrumentation have no unit or integration test coverage. The existing integration test base (`WebApiIntegrationTestBase`) should at minimum be checked to ensure it boots correctly with the new OTel middleware in place (the tenant-scope middleware at line 286-301 of `SharedConfiguration.cs` now sits in the middleware pipeline unconditionally, which could affect test harness behaviour). Consider adding: (1) a unit test for `ResolveTenantTagName` priority logic; (2) a unit test for `MongoDbTenantPropagationProcessor.OnStart` propagation; (3) a unit test for the updated `WorkflowSignalService` constructor with the new `IConfiguration` parameter.

#### Performance

The `MongoDbTenantPropagationProcessor` walks the `Activity.Parent` chain on every MongoDB operation start. In applications with deeply nested call stacks this is a linear walk per DB call. Under load this could add measurable overhead. The `ClusterConfigurator` lambda in `MongoDbClientService` captures `excludedCommands` and `captureCommandText` from environment variables at client construction time (singleton scope), which is correct — no per-call environment reads. The OTLP exporter uses gRPC for both traces and metrics; ensure the target collector supports gRPC (not HTTP/JSON) as this is not the default for all OTEL Collector deployments.

---

### Files Reviewed

| `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs` | +230/-0 | High | Core OTel wiring, PII handling, new middleware |
| `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs` | +26/-0 | High | Middleware pipeline modified; logging scope bug |
| `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs` | +33/-7 | Medium | PII inconsistency on user.id tag; typo in log |
| `XiansAi.Server.Src/Shared/Data/MongoDbClientService.cs` | +25/-0 | Medium | Env var reads bypass IConfiguration |
| `XiansAi.Server.Src/Shared/Utils/Temporal/NewWorkflowOptions.cs` | +9/-8 | Low | AddToMemo may be unused dead code |
| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs` | +4/-0 | Low | TracingInterceptor registration; correct |
| `XiansAi.Server.Src/XiansAi.Server.csproj` | +11/-1 | Low | New OTel package references; BOM char removed |
| `XiansAi.Server.Src/.env.example` | +22/-0 | Low | Good documentation of new env vars |

XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs
```diff
@@ -274,6 +278,37 @@ public static WebApplication UseSharedMiddleware(this WebApplication app)
         app.UseAuthentication();
         app.UseAuthorization();
 
+        // After auth resolves the tenant, push the tenant tag into the logging scope so every
+        // log record emitted within a request carries it as a structured attribute.
+        var tenantTagName = OpenTelemetryExtensions.ResolveTenantTagName(app.Configuration);
+        var scopeLogger = app.Services.GetRequiredService<ILoggerFactory>()
+            .CreateLogger("XiansAi.TenantScope");
+
+        app.Use(async (context, next) =>
```
xianix-agent: **Warning:** The tenant-scope middleware and `BeginScope` dictionary allocation run on every request even when `OpenTelemetry:Enabled=false`. When OTel is disabled, `Activity.Current` is always `null`, so the scope is only useful if another structured logging sink consumes it. Add a comment clarifying this is intentional for structured logging, or gate the middleware registration with the `Enabled` flag to avoid the per-request allocation on non-OTel deployments.

XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs
```diff
@@ -77,26 +84,51 @@ public class WorkflowSignalService : IWorkflowSignalService
...
+            {
+                activity?.SetTag(_tenantTagName, _tenantContext.TenantId);
+            }
+
+            if (_includeUserIdentity && !string.IsNullOrEmpty(_tenantContext.LoggedInUser))
+            {
+                activity?.SetTag("user.id", _tenantContext.LoggedInUser);
+            }
+
             var client = await _clientFactory.GetClientAsync() ?? throw new Exception("Failed to get Temporal client");
```
xianix-agent: **Warning:** Throwing a base `Exception` makes it impossible for callers or middleware to distinguish a Temporal connectivity failure from other error categories. Use `InvalidOperationException` (or a domain-specific exception type) instead:

```csharp
var client = await _clientFactory.GetClientAsync()
    ?? throw new InvalidOperationException("Temporal client factory returned null — check Temporal configuration.");
```

XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs
```diff
@@ -77,26 +84,51 @@ public class WorkflowSignalService : IWorkflowSignalService
...
 
     public async Task<IResult> SignalWithStartWorkflow(WorkflowSignalWithStartRequest request)
     {
+        using var activity = ActivitySource.StartActivity("Temporal.SignalWithStart", ActivityKind.Client);
+
         try
         {
+            activity?.SetTag("temporal.operation", "signal_with_start");
+            activity?.SetTag("temporal.workflow_type", request.TargetWorkflowType);
+            activity?.SetTag("temporal.signal_name", request.SignalName ?? "unknown");
```
xianix-agent: **Suggestion:** `request.SignalName ?? "unknown"` sets a span tag that says `unknown`, but line 143 later throws when `SignalName` is null. Validate `SignalName` before setting tags so the span never carries a misleading value:

```csharp
if (request.SignalName == null) throw new ArgumentException("SignalName is required", nameof(request));
// then set span tags
```

xianix-agent (changes requested):
## PR Review Report

**PR:** OpenTelemetry implementation (#388)
**Branch:** `Otel-fixes`
**Author:** Isanka Patabandige <isankap@99x.io>
**Files Changed:** 8 | **+387** / **-10**

---

### Summary

This PR introduces a solid OpenTelemetry observability layer covering tracing, metrics, and structured logging across the ASP.NET Core request pipeline, MongoDB operations, and Temporal workflow signaling. The PII safeguards (user identity opt-in, command text opt-in) are well-considered and clearly documented. However, there are several correctness and design issues that should be addressed before merge: the `MongoDbClientService` always installs the diagnostics subscriber regardless of whether OTel is enabled, the `_memo` backing field is declared but never actually used for mutation (making it dead code in the current state), and the `NewWorkflowOptions` constructor still accepts an `idPostfix` parameter named differently from the `WorkflowOptions.Id` parameter (minor but confusing). Additionally, there are no unit tests for any of the new OTel logic.

---

- [ ] `XiansAi.Server.Src/Shared/Data/MongoDbClientService.cs:101` — `DiagnosticsActivityEventSubscriber` is always registered, even when `OpenTelemetry:Enabled=false`. This adds unnecessary overhead (event subscription, hash-set lookups on every MongoDB command) in environments that have not opted into telemetry. The subscriber should only be attached when OTel is enabled, or the subscriber registration should be guarded by the same `OpenTelemetry:Enabled` flag used in `OpenTelemetryExtensions`.

  ```csharp
  // Current (problematic) — subscriber always installed
  settings.ClusterConfigurator = cb =>
      cb.Subscribe(new DiagnosticsActivityEventSubscriber(new InstrumentationOptions { ... }));

  // Fix — guard with the enabled flag
  var otelEnabled = _configuration.GetValue<bool>("OpenTelemetry:Enabled", false);
  if (otelEnabled)
  {
      settings.ClusterConfigurator = cb =>
          cb.Subscribe(new DiagnosticsActivityEventSubscriber(new InstrumentationOptions { ... }));
  }
  ```

- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/NewWorkflowOptions.cs:12` — `_memo` is declared as a `private readonly Dictionary<string, object>` backing field with a comment saying it exists "so `AddToMemo` can always mutate it directly", but no `AddToMemo` method exists anywhere in the file or in the diff. The backing field is assigned and then immediately assigned back to the base `Memo` property — if the base class replaces `Memo` with a copy (which `WorkflowOptions` likely does on serialization), mutations via `_memo` will silently be ignored. Either add the `AddToMemo` method this field was intended for, or remove the backing field and inline the dictionary directly into `Memo = GetMemo(...)` as before.

  ```csharp
  // Current (dead code / misleading intent)
  private readonly Dictionary<string, object> _memo;
  // ...
  _memo = GetMemo(...);
  Memo = _memo;

  // Fix option A — remove backing field if AddToMemo is not needed
  Memo = GetMemo(tenantContext, agentName, systemScoped, effectiveUserId, idPostfix ?? string.Empty);

  // Fix option B — implement the advertised AddToMemo helper
  public void AddToMemo(string key, object value) => _memo[key] = value;
  ```

---

### Warnings (Should Fix)

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:87` — A short-lived `LoggerFactory` is created with `using var` just to emit a handful of startup messages. This factory is disposed before `builder.Build()` is called, meaning its logger cannot be used after the early-return branches. While this works now (the logger is only used within the same block), it is fragile: if a future developer adds logging after `builder.Services.AddOpenTelemetry()`, the logger will be disposed and the message will be silently dropped. Consider using `builder.Services.BuildServiceProvider()` to resolve a real logger, or simply use `Console.WriteLine` / `Debug.WriteLine` for one-time bootstrap messages, or store the log messages and re-emit them post-build.

- [ ] `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs:283` — `OpenTelemetryExtensions.ResolveTenantTagName(app.Configuration)` is called unconditionally in `UseSharedMiddleware`, meaning the tenant-scope middleware is registered even when `OpenTelemetry:Enabled=false`. When OTel is disabled, `Activity.Current` will always be `null`, so `Activity.Current?.GetTagItem(...)` is always `null`, and the code falls back to `ITenantContext`. This path works correctly but allocates a logger scope dictionary on every request for no observability benefit. The middleware should be guarded with the same `Enabled` check, or the comment should clarify this is intentional for structured logging even without OTel export.

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:127` — Generic `Exception` is used rather than a more descriptive custom exception type when the Temporal client cannot be obtained. This makes it harder to distinguish a configuration/connectivity failure from a logic error at the call site.

  ```csharp
  // Current
  var client = await _clientFactory.GetClientAsync() ?? throw new Exception("Failed to get Temporal client");

  // Fix
  var client = await _clientFactory.GetClientAsync()
      ?? throw new InvalidOperationException("Temporal client factory returned null — check Temporal configuration.");
  ```

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:88` — The bootstrap logger is configured with `AddSimpleConsole()` only. In containerised environments (Docker/Kubernetes) the default log collector usually expects JSON output. If the application is already configured with a JSON formatter (e.g. `AddJsonConsole`), the startup OTel messages will be emitted as plain text and break structured log aggregation. Match the format used by the rest of the application, or emit startup OTel messages through the DI logger after `app.Build()`.

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:112` — `activity.SetTag("http.status_code", ...)` duplicates the semantic convention tag that `AddAspNetCoreInstrumentation` already emits as `http.response.status_code` (the new OTel HTTP semconv). This will produce a redundant `http.status_code` attribute alongside the framework-emitted `http.response.status_code`, polluting every span with a duplicate. Remove the manual `SetTag` in `EnrichWithHttpResponse`.

  ```csharp
  // Current (adds redundant tag)
  options.EnrichWithHttpResponse = (activity, httpResponse) =>
  {
      activity.SetTag("http.status_code", httpResponse.StatusCode);  // remove this line
      ...
  };
  ```

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:128` — `activity.SetTag("http.request.method", request.Method.ToString())` in `EnrichWithHttpRequestMessage` also duplicates what the HTTP client instrumentation already captures under `http.request.method`. Remove the manual tag to avoid duplication.

---

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:195` — `MongoDbTenantPropagationProcessor` only overrides `OnStart`. Since `BaseProcessor<Activity>` is a no-op for `OnEnd`, this is fine, but an XML `<summary>` doc comment explaining that `OnEnd` is intentionally omitted would help future maintainers who might wonder if it is forgotten.

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:182` — `AddSource("XiansAi.*")` uses a wildcard prefix. This matches any `ActivitySource` whose name begins with `XiansAi.`, including future ones. This is likely intentional, but it is worth documenting explicitly so teams know new `ActivitySource` names must follow the `XiansAi.*` naming convention to be automatically captured.

- [ ] `XiansAi.Server.Src/XiansAi.Server.csproj:590` — All OpenTelemetry packages are pinned to `1.10.0`. Consider pinning via a `<PackageVersion>` in a central `Directory.Packages.props` to avoid version drift across OTel packages if the project adopts Central Package Management in the future.

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:113` — `request.SignalName ?? "unknown"` silently substitutes `"unknown"` in the span tag, but line 143 later throws `Exception("SignalName is required")` when it is null. Validate `SignalName` early (before setting span tags) so the null check is consistent and the span does not carry a misleading `"unknown"` value before throwing.

- [ ] `XiansAi.Server.Src/.env.example:131` — The commented-out env var key `OPENTELEMETRY_TENANT_TAG_NAME` uses `SCREAMING_SNAKE_CASE` while the primary configuration key uses `OpenTelemetry__TenantTagName` (double-underscore ASP.NET Core convention). The dual-key fallback is handled correctly in code, but the `.env.example` should clarify that the `OPENTELEMETRY_TENANT_TAG_NAME` key is the legacy/alternate form, with `OpenTelemetry__TenantTagName` being preferred for new deployments.

---

### Review Details

#### Code Quality

The new `OpenTelemetryExtensions.cs` is well-structured: clear separation of tracing, metrics, and logging configuration; defensive null checks; and a catch-all that prevents OTel init failures from crashing the server. The `MongoDbTenantPropagationProcessor` is clean and correctly walks the parent chain. The fix to `WorkflowSignalService` from `.Result` (sync-over-async) to `await` is a positive side-effect improvement in this PR. The `_memo` backing field in `NewWorkflowOptions` represents incomplete refactoring and should be resolved. The dual-key fallback pattern (config key + env var) is thoughtful for backward compatibility but leads to scattered configuration resolution logic — a centralized `OpenTelemetryOptions` POCO resolved via `IOptions<T>` would make this cleaner in the long run.

#### Security

PII handling is commendable: user identity fields, MongoDB command text, and user roles are all opt-in with off-by-default guards. The `IncludeUserIdentity` flag is checked consistently in both `OpenTelemetryExtensions` and `WorkflowSignalService`. The X-Tenant-Id header is read directly from the incoming request and set as a span tag — this is low-risk since it is informational only and the authenticated tenant ID from `ITenantContext` takes precedence in `EnrichWithHttpResponse`. No secrets are leaked through spans or logs. No new authentication surfaces are introduced.

#### Test Coverage

No tests were added for any of the new OTel logic. The following scenarios should have unit test coverage:

1. `ResolveTenantTagName` — returns default `tenant.id`, respects config key, respects env var, correct priority order.
2. `AddOpenTelemetry` — returns early when disabled, returns early when endpoint is empty.
3. `MongoDbTenantPropagationProcessor.OnStart` — propagates tag from immediate parent, propagates from grandparent, no-ops on non-MongoDB spans, no-ops when no ancestor has the tag.
4. `WorkflowSignalService` — activity tags set correctly, PII guard respected when `IncludeUserIdentity=false`.
5. `MongoDbClientService` — subscriber not attached when OTel disabled (once the guard is added).

#### Performance

The `MongoDbTenantPropagationProcessor` walks the parent chain on every MongoDB span start. In deeply nested call stacks this is O(depth) per MongoDB operation, but in practice HTTP request depths are shallow (2–4 levels) so this is negligible. The `excludedCommands` HashSet construction in `MongoDbClientService.CreateMongoClient` happens once at startup — correctly placed. The `scopeLogger.BeginScope(new Dictionary<string, object> {...})` in the tenant middleware allocates a new dictionary on every request; this could be micro-optimized with a cached `IReadOnlyList<KeyValuePair<string, object>>`, but is unlikely to be a bottleneck. The `GetClientAsync` synchronization uses a single `SemaphoreSlim(1,1)` globally, which could serialize client creation across all tenants simultaneously connecting — this is pre-existing and outside the scope of this PR but worth a follow-up.

---

### Files Reviewed

| `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs` | +227 | Medium | New file; core OTel wiring; duplicate semantic convention tags |
| `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs` | +35 | Medium | Middleware always registered regardless of OTel enabled flag |
| `XiansAi.Server.Src/Shared/Data/MongoDbClientService.cs` | +38 | Medium | Subscriber always installed; should be guarded |
| `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs` | +44 | Low | Good async fix; minor exception type concern |
| `XiansAi.Server.Src/Shared/Utils/Temporal/NewWorkflowOptions.cs` | +11 | Low | Dead backing field; misleading comment |
| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs` | +4 | Low | `TracingInterceptor` always added; safe per comment |
| `XiansAi.Server.Src/XiansAi.Server.csproj` | +12 | Low | 7 new OTel package references; all pinned consistently |
| `XiansAi.Server.Src/.env.example` | +26 | Low | Clear documentation; minor naming convention inconsistency |

---

*Review generated by Claude Code — claude-sonnet-4-6*

xianix-agent (changes requested):
## PR Review Report

**PR:** Add OpenTelemetry support for tracing and metrics (branch: pr-388-1618)
**Author:** Isanka Patabandige <isankap@99x.io>
**Files Changed:** 8 | **+387** / **-10**

---

### Summary

This PR introduces a well-structured, opt-in OpenTelemetry integration covering tracing, metrics, and logs with appropriate PII guards for tenant identity and user fields. The overall design is solid — fail-safe initialization, sensible defaults, and dual configuration-key backwards compatibility are all appreciated. However, there are several correctness issues that must be addressed before merge: the tenant-scope middleware is registered after `UseAuthorization` in the pipeline, which will cause it to be silently bypassed for requests that hit Minimal API endpoints (which short-circuit before custom middleware), and a separate timing issue means the `ActivitySource` declared in `WorkflowSignalService` will create spans that are never exported unless a `TracerProvider` has been activated for it first. There are also a handful of warnings around resource disposal, duplicated URI construction, and test coverage gaps.

---

- [ ] `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs:281` — **Tenant-scope middleware registered after `UseAuthorization`; it will be bypassed by Minimal API endpoints.**

  In ASP.NET Core, Minimal API endpoint handlers execute _inside_ the routing/endpoint middleware, which comes after `UseAuthorization`. Any `app.Use(...)` middleware added _after_ `UseAuthorization` is inserted into the pipeline after auth but still before the terminal endpoint execution — however, because Minimal API endpoints use `MapXxx` and short-circuit through `UseEndpoints` internally, the custom `app.Use` middleware added here is placed _after_ `UseAuthorization` but _before_ the endpoint middleware, which means it should run. That said, the canonical and safe position for cross-cutting tenant enrichment middleware is **between `UseAuthentication` and `UseAuthorization`**, not after both. As written, if the endpoint short-circuits before reaching this middleware (e.g., through a middleware that calls `context.Response.CompleteAsync()` early, or a future refactor), the scope will be lost. More critically: because `app.Use(...)` is called _conditionally at startup_ while `UseAuthentication`/`UseAuthorization` are called unconditionally, the ordering cannot be easily verified by inspection. Move the middleware registration to between `UseAuthentication` and `UseAuthorization`:

  ```csharp
  // Current (risky ordering — tenant scope middleware appended after auth)
  app.UseAuthentication();
  app.UseAuthorization();

  if (app.Configuration.GetValue<bool>("OpenTelemetry:Enabled", false))
  {
      app.Use(async (context, next) => { ... });
  }

  // Fix — register between auth stages so it always runs for authenticated requests
  app.UseAuthentication();

  if (app.Configuration.GetValue<bool>("OpenTelemetry:Enabled", false))
  {
      app.Use(async (context, next) => { ... });
  }

  app.UseAuthorization();
  ```

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:47-49` — **Bootstrap `LoggerFactory` disposed before `AddOpenTelemetry` builder lambdas execute.**

  `startupLoggerFactory` is wrapped in a `using` block, meaning it is disposed at the end of the `AddOpenTelemetry` method. However, `builder.Services.AddOpenTelemetry()` registers lambda callbacks that are not executed until the `WebApplication` is built (i.e., after this method returns). The `logger` variable captured inside those lambdas (e.g., the final `logger.LogInformation(...)` call on line 173) references the already-disposed factory. While the call on line 173 happens inside `try` (before the method returns), it is the only log statement that is safe. All statements are technically within the `try` block scope, so disposal happens after them — but this is fragile and will cause issues if any log call is added inside the lambda callbacks registered to `ConfigureResource`, `WithTracing`, or `WithMetrics`. The logger should not be disposed until after all setup completes. Alternatively, use `ILogger<OpenTelemetryExtensions>` from the DI container, or make the `using` scope narrower to just the pre-registration logging section.

- [ ] `XiansAi.Server.Src/Shared/Data/MongoDbClientService.cs:101-110` — **`DiagnosticsActivityEventSubscriber` registered unconditionally at `MongoClient` creation time, but `ClusterConfigurator` is only a guard on `otelEnabled`. The `excludedCommands` and `captureCommandText` variables are evaluated even when OTel is disabled, which is harmless but inconsistent. The real bug: if `MongoDbClientService` is registered as a singleton (which is typical for expensive clients), toggling `OpenTelemetry:Enabled` at runtime has no effect because the subscriber is baked in at construction time.** This is an architectural note, not a crash bug, but it means documentation should clarify that changing `OpenTelemetry:Enabled` requires an application restart to take effect — add a comment to the `.env.example` and the startup log.

---

### Warnings (Should Fix)

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:143-147` and `153-157` — **OTLP exporter `Uri` constructed twice (once for tracing, once for metrics), and again for logging (line 168-170).** Extract the URI construction into a local variable to avoid repeated allocations and to make it clear a single endpoint is used for all signals:

  ```csharp
  // Current
  .AddOtlpExporter(options =>
  {
      options.Endpoint = new Uri(otlpEndpoint);
      options.Protocol = OpenTelemetry.Exporter.OtlpExportProtocol.Grpc;
  }))
  // ... repeated two more times

  // Fix
  var otlpUri = new Uri(otlpEndpoint);
  // Use otlpUri in all three AddOtlpExporter calls
  ```

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:71` — **`ActivitySource` is static but the source name `"XiansAi.Server.Temporal"` must be registered via `.AddSource("XiansAi.Server.Temporal")` (or the wildcard `"XiansAi.*"`) in the tracer configuration before any spans will be exported.** The wildcard `.AddSource("XiansAi.*")` on line 140 of `OpenTelemetryExtensions.cs` does cover this, but only when OpenTelemetry is enabled. When OTel is disabled, `ActivitySource.StartActivity(...)` returns `null` due to no listeners — this is correct behavior. However, the source is declared as `static readonly`, meaning it is created at class load time regardless of whether OTel is configured. This is fine by convention, but document it with a comment so future maintainers understand why `activity` can legitimately be null.

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:127` — **`throw new Exception(...)` should use a more specific exception type.** `throw new Exception("Failed to get Temporal client")` is a generic exception that will be caught by the broad `catch (Exception ex)` handler below, which then rethrows. Use `InvalidOperationException` for predictable failure states:

  ```csharp
  // Current
  var client = await _clientFactory.GetClientAsync() ?? throw new Exception("Failed to get Temporal client");

  // Fix
  var client = await _clientFactory.GetClientAsync() ?? throw new InvalidOperationException("Failed to get Temporal client");
  ```

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:143` — **`throw new Exception("SignalName is required")` is also generic.** Use `ArgumentException` or `InvalidOperationException`:

  ```csharp
  // Current
  if (request.SignalName == null) throw new Exception("SignalName is required");

  // Fix
  if (request.SignalName == null) throw new ArgumentException("SignalName is required", nameof(request));
  ```

- [ ] `XiansAi.Server.Src/XiansAi.Server.csproj:49-51` — **Test-only packages (`Moq`, `xunit`) are listed in the main project file rather than in a test project file.** These packages should not be included in production assemblies. Move them to `XiansAi.Server.Tests.csproj` or add `<PrivateAssets>all</PrivateAssets><IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>` conditionally. (Note: this is a pre-existing issue, but the PR adds more packages to the same file, making it a good opportunity to fix.)

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:88-89` — **Blank lines inside lambda bodies (lines 88-89 and 99-100 and 125-126) are style noise.** The OpenTelemetry enrichment lambdas have trailing blank lines before the first statement. These should be removed for consistency with the surrounding code style.

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:203-222` — **`MongoDbTenantPropagationProcessor.OnStart` walks the parent chain linearly at span start time.** While this is functionally correct, deeply nested activity trees (e.g., inside retry loops or parallel operations) could traverse many hops. Consider capping traversal depth or caching the resolved value. For the current codebase depth this is acceptable, but worth a comment:

  ```csharp
  // Consider adding a depth cap to avoid O(depth) traversal in pathological cases:
  const int MaxDepth = 10;
  int depth = 0;
  var parent = activity.Parent;
  while (parent != null && depth++ < MaxDepth)
  ```

---

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:36` — `typeof(Program).Assembly.GetName().Version` relies on `Program` being in the same assembly as the server. This works today, but consider using `Assembly.GetEntryAssembly()` for a more robust approach if the code is ever extracted to a shared library.

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:22-25` — The `ResolveTenantTagName` method reads from `Environment.GetEnvironmentVariable` as a fallback. ASP.NET Core's `IConfiguration` already integrates environment variables by default (via `ASPNETCORE_` prefix or full env var binding). Reading raw env vars alongside `IConfiguration` creates two code paths that can diverge. Consider exposing this only through `IConfiguration` to keep configuration resolution consistent. At minimum, document why the raw env var fallback is necessary alongside the configuration key.

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:111-115` — Tags are set on the activity before the try block's main work, which is fine. However, `temporal.workflow_id` is logged as `"auto-generated"` when `TargetWorkflowId` is null, but the actual workflow ID is computed inside `NewWorkflowOptions` from tenant context. Consider updating the tag after the options are created to reflect the resolved ID.

- [ ] General — No unit tests or integration tests were added for the new OpenTelemetry configuration logic. At minimum, unit tests for `ResolveTenantTagName` (covering configuration override, env var fallback, and default) and `MongoDbTenantPropagationProcessor` (verifying tenant tag propagation from parent to child span) would improve confidence in the feature's correctness and prevent regressions.

- [ ] `XiansAi.Server.Src/.env.example:131` — The `OPENTELEMETRY_TENANT_TAG_NAME` environment variable is documented in `.env.example` as a raw env var (not using the `__` double-underscore ASP.NET Core convention), while the primary key `OpenTelemetry__TenantTagName` is absent from the example file. Add the primary key to the example for discoverability, alongside the legacy env var.

---

### Review Details

#### Code Quality

The implementation is clean and well-commented. Naming is consistent, the opt-in pattern for PII fields is correctly applied throughout all enrichment points (HTTP request, HTTP response, Temporal span, MongoDB propagation), and the `MongoDbTenantPropagationProcessor` correctly addresses the tag inheritance gap in the MongoDB driver. The fix of `.Result` to `await` in `WorkflowSignalService` (line 131) is a meaningful correctness improvement that prevents thread-pool starvation. The `NewWorkflowOptions` refactor (direct dictionary initialization) is a clean simplification. The startup logger pattern is thoughtful. Main concern is the middleware ordering issue noted above and the use of generic `Exception` types in two places.

#### Security

PII handling is the most security-sensitive aspect of this PR and is handled responsibly: user identity fields (`tenant.user`, `tenant.user_type`, `tenant.user_roles`) are off by default and clearly documented. MongoDB `CaptureCommandText` is off by default, which prevents query filter values from leaking into traces. The `X-Tenant-Id` header is read from incoming requests — ensure that this header is only trusted from authenticated/validated sources (i.e., not blindly accepted from untrusted clients). The current code does not validate or sanitize `tenantId` read from the header before setting it as a span tag; extremely long values could be used to inflate span attribute storage. Consider capping the length. No secrets or credentials are logged or exported.

#### Test Coverage

No tests were added for any of the new code paths. The following scenarios are unverified by automated tests:
- `ResolveTenantTagName` with each of its three resolution paths
- `MongoDbTenantPropagationProcessor` tenant tag propagation
- `WorkflowSignalService` activity creation and tag setting
- The middleware tenant scope registration when OTel is enabled vs disabled
- MongoDB `DiagnosticsActivityEventSubscriber` registration when `otelEnabled=false`

The existing integration tests do exercise `WorkflowSignalService` indirectly, but the new OTel instrumentation paths are not validated.

#### Performance

The `AddOtlpExporter` lambda is invoked three times with identical configuration, creating three `Uri` objects from the same string. This is micro-level but worth consolidating. The `MongoDbTenantPropagationProcessor` traverses the parent activity chain on every MongoDB span start — for the typical request depth this is O(1-3) hops, which is negligible. The startup bootstrap `LoggerFactory` is created and disposed synchronously at startup — this is acceptable and will not affect runtime performance. The `TracingInterceptor` added to all Temporal clients is noted as a no-op when no `TracerProvider` is active, which is the correct behavior per the SDK docs.

---

### Files Reviewed

| `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs` | +225 | High | Core OTel wiring; middleware ordering issue |
| `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs` | +38 | High | Middleware registration ordering concern |
| `XiansAi.Server.Src/Shared/Data/MongoDbClientService.cs` | +42 | Medium | OTel MongoDB instrumentation; IConfiguration injection |
| `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs` | +44 | Medium | Span creation; async fix; generic exception types |
| `XiansAi.Server.Src/XiansAi.Server.csproj` | +12 | Low | New OTel package references; test packages in prod project |
| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs` | +4 | Low | `TracingInterceptor` added cleanly |
| `XiansAi.Server.Src/Shared/Utils/Temporal/NewWorkflowOptions.cs` | -4 | Low | Minor refactor; no logic change |
| `XiansAi.Server.Src/.env.example` | +26 | Low | Documentation; well-commented |

xianix-agent (changes requested):
## PR Review Report

**PR:** Add OpenTelemetry support for tracing and metrics (branch: `pr-388-2080`)
**Author:** Isanka Patabandige &lt;isankap@99x.io&gt;
**Files Changed:** 8 | **+387** / **-10**

---

### Summary

This PR introduces comprehensive OpenTelemetry observability support — tracing, metrics, and structured logging — across the XiansAi.Server application. The implementation is feature-complete and follows secure-by-default patterns for PII (user identity and MongoDB command text are both opt-in). There are several issues that must be resolved before merge: test packages are co-mingled in the production project file, the singleton startup-time constraint for MongoDB instrumentation is undocumented, and a handful of code quality and robustness issues need attention.

---

- [ ] `XiansAi.Server.Src/XiansAi.Server.csproj:49-50` — **Test packages (`Moq` v4.20.72 and `xunit` v2.9.3) are declared in the production project file, not the test project.** These libraries will be bundled into the production Docker image, increasing the attack surface and binary size unnecessarily. Remove both `PackageReference` entries from `XiansAi.Server.csproj` and confirm they exist only in `XiansAi.Server.Tests.csproj`.

  ```xml
  <!-- REMOVE from XiansAi.Server.csproj (production project): -->
  <PackageReference Include="Moq" Version="4.20.72" />
  <PackageReference Include="xunit" Version="2.9.3" />
  ```

- [ ] `XiansAi.Server.Src/Shared/Data/MongoDbClientService.cs:84-110` — **The `DiagnosticsActivityEventSubscriber` is registered once at construction time, and `MongoDbClientService` is a singleton.** Any runtime change to `OpenTelemetry:Enabled`, `ExcludedCommands`, or `CaptureCommandText` (e.g. via Azure App Configuration hot-reload) will be silently ignored. This is acceptable, but the constraint is undocumented, making it a latent maintenance hazard. Add an explicit comment:

  ```csharp
  // NOTE: MongoDB instrumentation is evaluated once at startup (MongoDbClientService is a singleton).
  // Changes to OpenTelemetry:Enabled or MongoDB sub-settings at runtime require an application restart.
  var otelEnabled = _configuration.GetValue<bool>("OpenTelemetry:Enabled", false);
  ```

- [ ] `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs:281-313` — **The tenant-scope middleware's fallback path (`ITenantContext.TenantId`) does not log when a tenant ID cannot be resolved.** For mTLS/API-key paths that authenticate without an `X-Tenant-Id` header and where auth has failed (401/403), spans and log records will carry no tenant tag with no trace of why. Add a debug-level log so the absence of tenant context is observable:

  ```csharp
  if (string.IsNullOrWhiteSpace(tenantId))
  {
      scopeLogger.LogDebug("No tenant context available for request {Path}", context.Request.Path);
      await next(context);
      return;
  }
  ```

---

### Warnings (Should Fix)

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:143` — `throw new Exception("SignalName is required")` uses the base `Exception` type. Use `ArgumentException` to correctly classify this as a caller error:

  ```csharp
  if (request.SignalName == null)
      throw new ArgumentException("SignalName is required", nameof(request));
  ```

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:127` — Similarly, `throw new Exception("Failed to get Temporal client")` should use `InvalidOperationException` for better diagnosability.

- [ ] `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs:303` — `new Dictionary<string, object> { [tenantTagName] = tenantId }` allocates a dictionary on every HTTP request in the hot-path middleware. For high-throughput scenarios, prefer a `KeyValuePair[]` which the logging infrastructure accepts directly and does not require a dictionary heap allocation:

  ```csharp
  using (scopeLogger.BeginScope(new[] { new KeyValuePair<string, object>(tenantTagName, tenantId!) }))
  {
      await next(context);
  }
  ```

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:36` — `typeof(Program).Assembly.GetName().Version` returns null in single-file publish mode; the fallback `?? "1.0.0"` is misleading in production. Use `AssemblyInformationalVersionAttribute` which is always populated by the .NET SDK:

  ```csharp
  var serviceVersion =
      typeof(Program).Assembly
          .GetCustomAttribute<System.Reflection.AssemblyInformationalVersionAttribute>()
          ?.InformationalVersion
      ?? typeof(Program).Assembly.GetName().Version?.ToString()
      ?? "unknown";
  ```

- [ ] `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs:71` — `ActivitySource` is `static readonly` but is never disposed. While the runtime will clean it up on application exit, add a comment acknowledging the intentional choice to prevent future developers from adding unnecessary disposal logic:

  ```csharp
  // ActivitySource lifetime matches the application lifetime; no explicit disposal required.
  private static readonly ActivitySource ActivitySource = new("XiansAi.Server.Temporal");
  ```

---

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:27` — This `AddOpenTelemetry(this WebApplicationBuilder builder)` extension has the same name as the one provided by `OpenTelemetry.Extensions.Hosting`. While the overload resolves correctly today, the identical name can confuse developers about which implementation is active. Consider renaming to `AddXiansAiOpenTelemetry` or `AddObservability`.

- [ ] `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs:203-222` — `MongoDbTenantPropagationProcessor.OnStart` walks the full Activity parent chain on every MongoDB operation. In deep call stacks this is O(depth). Consider caching the resolved tenant ID in an `AsyncLocal<string?>` (set by the tenant-scope middleware) to make propagation O(1) for high-throughput MongoDB workloads.

- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs:68` — `TracingInterceptor` is unconditionally added. The SDK comment notes it is a no-op without a registered `TracerProvider`, but verify whether it still injects/reads W3C trace-context headers on every gRPC call even when OTel is disabled. If there is any overhead, condition it on the `OpenTelemetry:Enabled` flag.

- [ ] No unit tests were added for `OpenTelemetryExtensions`, `MongoDbTenantPropagationProcessor`, or the OTel enrichment paths in `WorkflowSignalService`. Given the PII-sensitive opt-in nature of `IncludeUserIdentity`, a unit test asserting that user identity tags are NOT emitted by default would provide meaningful regression coverage. A test for `ResolveTenantTagName` (covering the config key, env-var fallback, and default) would also be straightforward to add.

- [ ] `XiansAi.Server.Src/.env.example` — Consider adding a note that setting `OpenTelemetry__Enabled=true` without a reachable OTLP endpoint is safe (the server logs a warning and continues without telemetry export) to reduce operator hesitation during initial configuration.

---

### Review Details

#### Code Quality

The implementation is well-structured and `OpenTelemetryExtensions` follows the single-responsibility principle. PII protection is handled thoughtfully with clear opt-in semantics and startup warnings. The `MongoDbTenantPropagationProcessor` inner class is correctly declared `private sealed`. A positive side-effect fix was included: `_agentService.IsSystemAgent(request.SourceAgent).Result` (blocking `.Result` on an async call, a common deadlock risk) was correctly replaced with `await _agentService.IsSystemAgent(request.SourceAgent)`. The blank lines inside lambda bodies (e.g. after `options.EnrichWithHttpRequest = (activity, httpRequest) => {`) are a minor style inconsistency worth cleaning up. The namespace `Features.Shared.Configuration` on a file at `Shared/Configuration/` is slightly unconventional but consistent with the existing project convention.

#### Security

PII handling is commendably conservative: user identity fields are opt-in with a startup warning, MongoDB command text is opt-in with documentation in `.env.example`. One low-severity observation: the `EnrichWithHttpRequest` callback tags the span with the raw `X-Tenant-Id` header value *before* authentication validates it. The `EnrichWithHttpResponse` callback correctly overwrites this with the validated `ITenantContext.TenantId` after auth, but the initial unvalidated value is briefly visible during the span lifetime. This is acceptable (the final tag is authoritative), but worth documenting. The test packages (`Moq`, `xunit`) in the production `.csproj` are a supply-chain hygiene concern and must be removed.

#### Test Coverage

No new tests were added for any of the introduced code paths. The existing integration test suite in `XiansAi.Server.Tests/` does not appear to include tests for `OpenTelemetryExtensions`, `MongoDbTenantPropagationProcessor`, or the new `WorkflowSignalService` OTel enrichment. The `ResolveTenantTagName` method and `IncludeUserIdentity` guard are the highest-priority targets for unit test coverage given their PII sensitivity.

#### Performance

The tenant-scope middleware allocates a `Dictionary<string, object>` per HTTP request for the logging scope — a `KeyValuePair[]` is preferred (see warning above). The `MongoDbTenantPropagationProcessor.OnStart` parent-chain walk is O(depth); for typical web request nesting of 2-3 levels this is negligible, but should be profiled under load. The OTLP exporter uses default batch export settings; for high-volume deployments consider tuning `BatchExportProcessorOptions.MaxQueueSize` and `ExporterTimeoutMilliseconds` to prevent back-pressure from a slow or unreachable collector.

---

### Files Reviewed

| `XiansAi.Server.Src/Shared/Configuration/OpenTelemetryExtensions.cs` | +225/-0 | High | Core OTel wiring; PII handling; processor inner class |
| `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs` | +37/-0 | Medium | Middleware registration order; per-request allocation |
| `XiansAi.Server.Src/Shared/Data/MongoDbClientService.cs` | +40/-2 | Medium | Singleton MongoDB instrumentation; startup-time only |
| `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs` | +42/-3 | Medium | Custom spans; PII opt-in; async bug fix (positive) |
| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs` | +4/-0 | Low | Always-on TracingInterceptor; safe as no-op |
| `XiansAi.Server.Src/Shared/Utils/Temporal/NewWorkflowOptions.cs` | +0/-4 | Low | Dead variable eliminated; clean refactor |
| `XiansAi.Server.Src/XiansAi.Server.csproj` | +11/-1 | Medium | Test packages in production project (must fix) |
| `XiansAi.Server.Src/.env.example` | +26/-0 | Low | Documentation; all defaults are secure |

PR #397, opened 2026-04-27
perf: Improving conversation message cycle time

### LINKED ISSUE #396: Improving conversation message cycle time
Coversations are initiated via adminapi where the UI sending a conversational request. Then its forwarded to the temporal as a signal. This signal is then delivered to the agent and agent calls the server via agentapi to fetch thread history. This is where the most of the time is spent. Once history is fetched, agent process and generate an oputput that is sent mach to server via agentapi. This response is saved to the database and mongodb streams process and trigger SSE channel to send the response back to UI. Lets look at the total process and see where we can improve the performance significantly at a lower risk.

hasith:
Closes #396

## Performance Report — Conversation Message Cycle Time

**Scope:** Full codebase (hot-path bias: adminapi → Temporal signal → agentapi thread-history → DB write → MongoDB change stream → SSE)
**Target runtime:** api
**Default branch:** main @ `1b0fa6b`
**Files in scope:** 196 source files (after excluding tests, docs, tooling, CI)
**Exclusions applied:** 110 files

---

### Analyzer Verdicts

| Analyzer | Status | Findings |
|---|---|---|
| latency-analyzer | OK | 6 |
| cpu-analyzer | OK | 4 |
| memory-analyzer | OK | 4 |
| io-query-analyzer | OK | 6 |

---

#### 1. PBKDF2 Key Derivation on Every Message Encrypt/Decrypt — **HIGH impact / HIGH confidence**
**Files:** `Shared/Services/SecureEncryptionService.cs:80-83`

`DeriveKey()` runs 200,000 PBKDF2-SHA256 iterations on every `Encrypt` and every `Decrypt` call. Since `(baseSecret, uniqueSecret)` is constant for the process lifetime, the 32-byte AES key is deterministic and safe to cache. On a history page of 50 messages: 50 × 200k = 10M SHA256 rounds on the request thread.

**Fix applied:** Static `ConcurrentDictionary` keyed by `(baseSecret, uniqueSecret)` — derivation runs at most once per key pair for the process lifetime.

---

#### 2. MongoDB Multi-Document Transaction on Every Message Save — **HIGH impact / HIGH confidence**
**Files:** `Shared/Repositories/ConversationRepository.cs:452-468`

Every message save (incoming + outgoing) opened a MongoDB multi-document transaction: `StartSession → BeginTx → InsertOne → UpdateOne → CommitTx` = 4 server round-trips. The `thread.UpdatedAt` field is cosmetic (UI sort order only) and does not need to be atomic with the message insert.

**Fix applied:** Replace transaction with direct `InsertOneAsync`; queue the `thread.UpdatedAt` update to the existing `BackgroundTaskService` (fire-and-forget). Removes 3 MongoDB round-trips per message on the hot path.

---

#### 3. Uncached Workflow-Type Validation — DB Hit on Every AdminAPI Call — **MEDIUM impact / HIGH confidence**
**Files:** `Shared/Services/ActivationValidationService.cs:76-80`

The activation entity lookup was cached (2-min TTL) but `ValidateWorkflowTypeForAgentAsync` was not — a fresh `flow_definition` DB query on every `/send`, `/listen`, `/history`, `/topics` adminapi call. Flow definitions rarely change at runtime.

**Fix applied:** Second cache entry keyed by `(tenantId, agentName, workflowType)` with 5-min TTL, using the existing `IAsyncResultCache` pattern.

---

#### 4. Two Sequential MongoDB Queries for Origin+Data on Outbound Messages — **HIGH impact / HIGH confidence**
**Files:** `Shared/Services/MessageService.cs:276-293`, `Shared/Repositories/ConversationRepository.cs`

`ProcessOutgoingMessage` issued two sequential `find + sort + limit 1` queries to fetch `origin` (for routing) and `data` (for Slack/Teams metadata) from the same collection with the same filter, differing only in projection.

**Fix applied:** New `GetLastIncomingOriginAndDataAsync` method projects both fields in a single round-trip; `MessageService` updated to call the merged method. Halves MongoDB I/O on every outbound app-origin message.

---

#### 5. Parallel SignalR Fan-out in MongoChangeStreamService — **MEDIUM impact / HIGH confidence**
**Files:** `Features/UserApi/Services/MongoChangeStreamService.cs`

For each change event, 2–3 `SignalR.SendAsync` calls fired sequentially, blocking processing of the next change event. Under load, a slow SignalR client stalls the entire change stream processor.

**Fix applied (previous commit):** Wrap all per-event `SendAsync` calls in `Task.WhenAll` so they execute concurrently instead of serially.

---

### Latency Risk Areas

| Finding | File | Impact | Confidence |
|---|---|---|---|
| L1: Sequential DB + Temporal signal in ProcessIncomingMessage | `Shared/Services/MessageService.cs:306-363` | High | High |
| L4: PBKDF2 KDF on every encrypt/decrypt | `Shared/Services/SecureEncryptionService.cs:80-83` | High | High |
| L6: MongoDB transaction per message save | `Shared/Repositories/ConversationRepository.cs:452-468` | High | High |
| L5: Sequential SignalR sends stall change-stream processor | `Features/UserApi/Services/MongoChangeStreamService.cs:118-207` | Medium | Medium |

---

### CPU & Memory Hotspots

| Finding | File | Impact | Confidence |
|---|---|---|---|
| C1: PBKDF2 200k iterations per call | `Shared/Services/SecureEncryptionService.cs` | High | High |
| C4: Unconditional JsonSerializer.Serialize in /converse logging | `Features/AgentApi/Endpoints/MessagingEndpoints.cs:172` | Low | Medium |
| M1/C3: Per-publish Delegate[] allocation in MessageEventPublisher | `Features/UserApi/Services/MessageEventPublisher.cs:59-66` | Medium | High |
| M4: Leaked LoggerFactory in static field | `Features/AgentApi/Endpoints/MessagingEndpoints.cs:26` | Low | High |

---

### I/O & Query Inefficiencies

| Finding | File | Impact | Confidence |
|---|---|---|---|
| Q1: MongoDB transaction per message save | `Shared/Repositories/ConversationRepository.cs:452-468` | High | High |
| Q2: Two separate queries for origin+data | `Shared/Services/MessageService.cs:276-293` | High | High |
| Q4: Uncached workflow-type validation | `Shared/Services/ActivationValidationService.cs:76-80` | Medium | High |
| Q5: ListCollectionNames on every reconnect | `Features/UserApi/Services/MongoChangeStreamService.cs:62-83` | Low | Low |
| Q3: Regex on WorkflowId in GetLastTaskIdAsync | `Shared/Repositories/ConversationRepository.cs:946-948` | Medium | High |

---

#### Quick Wins Applied in This PR (10 commits)

1. **`perf: cache IsSystemAgent result per process lifetime in WorkflowSignalService`** — eliminates repeated DB/RPC for agent system-scope check on every Temporal signal
2. **`perf: reuse static JsonSerializerOptions in SSEEventWriter`** — eliminates per-write allocations on the SSE flush path
3. **`perf: parallelise SignalR sends in MongoChangeStreamService with Task.WhenAll`** — change-stream processor no longer serialized by slow SignalR clients
4. **`perf: add two missing MongoDB indexes on conversation_message`** — `(workflow_id, participant_id, created_at)` and `(thread_id, created_at)` covering the two hottest query shapes
5. **`perf: cache PBKDF2-derived AES key`** — eliminates 200k-iteration KDF on every encrypt/decrypt call
6. **`perf: remove MongoDB transaction from SaveMessageAsync`** — 3 round-trips → 1; thread timestamp update moved to BackgroundTaskService
7. **`perf: cache workflow-type validation result (5-min TTL)`** — eliminates uncached DB hit on every adminapi send/listen/history/topics call
8. **`perf: merge GetLastIncomingOrigin + GetLastIncomingData into single query`** — halves MongoDB I/O on outbound app-origin messages
9. **`perf: replace unconditional JsonSerializer.Serialize in /converse logging`** — prevents full body serialization when log level is off
10. **`perf: skip redundant ListCollectionNames on MongoChangeStreamService reconnects`** — eliminates collection-existence check after first successful watch

#### Deeper Follow-Up (not applied — requires architectural discussion)

- **L1/Sequential ProcessIncomingMessage:** The `SaveMessage` write and `SignalWorkflowAsync` call are sequential. The signal already carries the message payload, so saving to MongoDB could be parallelized with the Temporal signal using `Task.WhenAll`. Requires verifying the agent does not read the DB for this specific message before the save completes.
- **C2/BSON conversion overhead:** `ConvertBsonToNativeObject` boxes every primitive into `object?` on every history fetch. Long-term, consider storing payload as serialized JSON string to avoid BSON ↔ .NET type-switching on reads.
- **M3/Unbounded BackgroundTaskService channel:** SingleReader means all DB writes are serialized. Under burst, the queue grows unbounded. Consider a bounded channel with configurable capacity and back-pressure.
- **Q3/Regex on WorkflowId:** `GetLastTaskIdAsync` uses a MongoDB `$regex` prefix scan that cannot use a B-tree index. Consider a compound index on `(base_workflow_id)` or storing the base workflow ID as a separate indexed field.
- **Q6/Blocking constructor in ConversationRepository:** `databaseService.GetDatabaseAsync().GetAwaiter().GetResult()` blocks the thread-pool during DI resolution. Migrate to `IHostedService` initialization pattern.
- **L5/SSE fan-out blocks change-stream:** Even with `Task.WhenAll`, SSE publishing runs inline in the change-stream loop. Long-term, a dedicated `Channel<MessageStreamEvent>` with separate consumer task would fully decouple SSE delivery latency from change-stream processing throughput.

---

### Files Assessed

196 runtime-critical C# source files across:
- `Features/AdminApi/` — HTTP endpoints, validation, SSE listen
- `Features/AgentApi/` — thread history, outbound message, activity history
- `Features/UserApi/` — MongoChangeStreamService, SSE stream handlers, MessageEventPublisher
- `Features/WebApi/` — workflow signaling, messaging service
- `Shared/Repositories/ConversationRepository.cs` — all message/thread DB operations
- `Shared/Services/` — MessageService, WorkflowSignalService, ActivationValidationService, SecureEncryptionService, PendingRequestService, BackgroundTaskService
- `Shared/Utils/Temporal/` — TemporalClientService, TemporalClientFactory
- `XiansAi.Server.Src/mongodb-indexes.yaml` — index definitions

110 files excluded: tests, docs, generated output, CI configs.

---

## Verification Checklist

- [ ] Run integration tests: `dotnet test XiansAi.Server.Tests/`
- [ ] Send a message via adminapi `/send` and verify it reaches the agent and SSE streams the response
- [ ] Verify agentapi `/history` decrypts messages correctly (PBKDF2 key cache validation)
- [ ] Check MongoDB logs — confirm no transaction entries for `conversation_message` inserts
- [ ] Verify `thread.UpdatedAt` is still updated (asynchronously) after message save
- [ ] Send 10 rapid messages and confirm change-stream processes them without backlog
- [ ] Confirm `/converse` endpoint still logs workflow/participant info at INFO level
- [ ] Load test: 50-message history page — measure response time before/after (expected: significant reduction from PBKDF2 cache)

### DIFF
--- a/XiansAi.Server.Src/Features/AgentApi/Endpoints/MessagingEndpoints.cs
+++ b/XiansAi.Server.Src/Features/AgentApi/Endpoints/MessagingEndpoints.cs
@@ -169,7 +169,11 @@ namespace Features.AgentApi.Endpoints
                 [FromQuery] int timeoutSeconds = 60,
                 [FromQuery] string? requestId = null) =>
             {
-                logger.LogInformation("Converse endpoint called with request: {Request}", JsonSerializer.Serialize(chatOrDataRequest));
+                if (logger.IsEnabled(LogLevel.Information))
+                {
+                    logger.LogInformation("Converse endpoint called for workflow {WorkflowId}, participant {ParticipantId}, type {Type}",
+                        chatOrDataRequest.WorkflowId, chatOrDataRequest.ParticipantId, type);
+                }
                 var workflow = chatOrDataRequest.WorkflowId;
                 if (string.IsNullOrEmpty(workflow))
                 {
--- a/XiansAi.Server.Src/Features/UserApi/Services/MongoChangeStreamService.cs
+++ b/XiansAi.Server.Src/Features/UserApi/Services/MongoChangeStreamService.cs
@@ -14,6 +14,11 @@ namespace Features.UserApi.Services
 {
     public class MongoChangeStreamService : BackgroundService
     {
+        // Skip the ListCollectionNames + CreateCollection roundtrip on reconnects after the
+        // first successful WatchAsync. The collection is created once at app startup and
+        // never dropped; re-checking on every transient-error reconnect adds unnecessary I/O.
+        private static volatile bool _collectionEnsured = false;
+
         private readonly IHubContext<ChatHub> _hubContext;
         private readonly IHubContext<TenantChatHub> _tenantHubContext;
         private readonly IServiceScopeFactory _scopeFactory;
@@ -67,19 +72,25 @@ namespace Features.UserApi.Services
                     var collectionName = "conversation_message";
                     var collection = database.GetCollection<ConversationMessage>(collectionName);
 
-                    // Ensure collection exists with retry logic
-                    var exists = await MongoRetryHelper.ExecuteWithRetryAsync(async () =>
+                    // Ensure collection exists — only on first iteration; subsequent reconnects skip this
+                    // to avoid a ListCollectionNames round-trip on every transient-error recovery.
+                    if (!_collectionEnsured)
                     {
-                        var collections = await database.ListCollectionNamesAsync(cancellationToken: stoppingToken);
-                        return await collections.ToListAsync(stoppingToken);
-                    }, _logger, maxRetries: 5, baseDelayMs: 1000, operationName: "ListCollectionNames");
-
-                    if (!exists.Contains(collectionName))
-                    {
-                        await MongoRetryHelper.ExecuteWithRetryAsync(async () =>
+                        var exists = await MongoRetryHelper.ExecuteWithRetryAsync(async () =>
                         {
-                            await database.CreateCollectionAsync(collectionName, cancellationToken: stoppingToken);
-                        }, _logger, maxRetries: 5, baseDelayMs: 1000, operationName: "CreateCollection");
+                            var collections = await database.ListCollectionNamesAsync(cancellationToken: stoppingToken);
+                            return await collections.ToListAsync(stoppingToken);
+                        }, _logger, maxRetries: 5, baseDelayMs: 1000, operationName: "ListCollectionNames");
+
+                        if (!exists.Contains(collectionName))
+                        {
+                            await MongoRetryHelper.ExecuteWithRetryAsync(async () =>
+                            {
+                                await database.CreateCollectionAsync(collectionName, cancellationToken: stoppingToken);
+                            }, _logger, maxRetries: 5, baseDelayMs: 1000, operationName: "CreateCollection");
+                        }
+
+                        _collectionEnsured = true;
                     }
 
                     var pipeline = new EmptyPipelineDefinition<ChangeStreamDocument<ConversationMessage>>()
--- a/XiansAi.Server.Src/Features/UserApi/Services/MongoChangeStreamService.cs
+++ b/XiansAi.Server.Src/Features/UserApi/Services/MongoChangeStreamService.cs
@@ -1,4 +1,4 @@
-﻿using Microsoft.AspNetCore.SignalR;
+using Microsoft.AspNetCore.SignalR;
 using MongoDB.Bson;
 using MongoDB.Driver;
 using Shared.Data;
@@ -38,7 +38,7 @@ namespace Features.UserApi.Services
             _tenantHubContext = tenantHubContext;
             _messageEventPublisher = messageEventPublisher ?? throw new ArgumentNullException(nameof(messageEventPublisher));
             _encryptionService = encryptionService ?? throw new ArgumentNullException(nameof(encryptionService));
-            
+
             // Get the unique secret for conversation messages
             _uniqueSecret = configuration["EncryptionKeys:UniqueSecrets:ConversationMessageKey"] ?? string.Empty;
             if (string.IsNullOrWhiteSpace(_uniqueSecret))
@@ -73,7 +73,7 @@ namespace Features.UserApi.Services
                         var collections = await database.ListCollectionNamesAsync(cancellationToken: stoppingToken);
                         return await collections.ToListAsync(stoppingToken);
                     }, _logger, maxRetries: 5, baseDelayMs: 1000, operationName: "ListCollectionNames");
-                    
+
                     if (!exists.Contains(collectionName))
                     {
                         await MongoRetryHelper.ExecuteWithRetryAsync(async () =>
@@ -102,11 +102,11 @@ namespace Features.UserApi.Services
                     // Wrap WatchAsync with retry logic to handle initial connection failures
                     using var cursor = await MongoRetryHelper.ExecuteWithRetryAsync(
                         async () => await collection.WatchAsync(pipeline, options, cancellationToken: stoppingToken),
-                        _logger, 
-                        maxRetries: 5, 
-                        baseDelayMs: 2000, 
+                        _logger,
+                        maxRetries: 5,
+                        baseDelayMs: 2000,
                         operationName: "WatchChangeStream");
-                    
+
                     // Iterate using MoveNextAsync and Current
                     while (await cursor.MoveNextAsync(stoppingToken))
                     {
@@ -115,7 +115,7 @@ namespace Features.UserApi.Services
                         var changeBatch = cursor.Current;
                         if (changeBatch == null) continue;
 
-                        foreach(var changeDoc in changeBatch) 
+                        foreach(var changeDoc in changeBatch)
                         {
                             if (stoppingToken.IsCancellationRequested) break;
 
@@ -124,7 +124,7 @@ namespace Features.UserApi.Services
 
                             // Convert BSON metadata to native .NET objects before sending via SignalR
                             ConvertBsonMetadataToObjectInternal(message);
-                            
+
                             // Decrypt the message text
                             DecryptMessageText(message);
 
@@ -136,7 +136,7 @@ namespace Features.UserApi.Services
                             {
                                 try
                                 {
-                                    _logger.LogDebug("Completing pending request {RequestId} with outgoing message {MessageId}", 
+                                    _logger.LogDebug("Completing pending request {RequestId} with outgoing message {MessageId}",
                                         message.RequestId, message.Id);
                                     pendingRequestService.CompleteRequest(message.RequestId, message, message.MessageType);
                                 }
@@ -149,44 +149,52 @@ namespace Features.UserApi.Services
                             // Existing SignalR broadcasting logic (updated for proper handoff routing)
                             if (message.Direction == MessageDirection.Outgoing || message.MessageType == MessageType.Handoff)
                             {
-                                // Route messages based on their type first, then content
+                                // perf: send all SignalR notifications for the same message concurrently
+                                // instead of awaiting each one sequentially. The sends are independent
+                                // (different clients/groups/methods) so no ordering constraint applies.
                                 if (message.MessageType == MessageType.Handoff)
                                 {
                                     // Handoff messages get their own dedicated event
                                     _logger.LogDebug("Sending handoff message to group {GroupId}: {Message}",
                                         groupId, JsonSerializer.Serialize(message));
-                                    await _hubContext.Clients.Group(groupId)
-                                        .SendAsync("ReceiveHandoff", message, cancellationToken: stoppingToken);
-                                    await _tenantHubContext.Clients.Group(tenantGroupId)
-                                        .SendAsync("ReceiveHandoff", message, cancellationToken: stoppingToken);
+                                    await Task.WhenAll(
+                                        _hubContext.Clients.Group(groupId)
+                                            .SendAsync("ReceiveHandoff", message, cancellationToken: stoppingToken),
+                                        _tenantHubContext.Clients.Group(tenantGroupId)
+                                            .SendAsync("ReceiveHandoff", message, cancellationToken: stoppingToken)
+                                    );
                                 }
                                 else if (message.MessageType == MessageType.Data)
                                 {
                                     // Data/Metadata messages (no text content)
                                     _logger.LogDebug("Sending metadata to group {GroupId}: {Message}",
                                         groupId, JsonSerializer.Serialize(message));
-                                    await _hubContext.Clients.Group(groupId)
+                                    await Task.WhenAll(
                                         // TODO: Remove the backward compatibility ReceiveMetadata later
-                                        .SendAsync("ReceiveMetadata", message, cancellationToken: stoppingToken);
-                                    await _hubContext.Clients.Group(groupId)
+                                        _hubContext.Clients.Group(groupId)
+                                            .SendAsync("ReceiveMetadata", message, cancellationToken: stoppingToken),
                                         // New method names
-                                        .SendAsync("ReceiveData", message, cancellationToken: stoppingToken);
-                                    await _tenantHubContext.Clients.Group(tenantGroupId)
-                                        .SendAsync("ReceiveData", message, cancellationToken: stoppingToken);
+                                        _hubContext.Clients.Group(groupId)
+                                            .SendAsync("ReceiveData", message, cancellationToken: stoppingToken),
+                                        _tenantHubContext.Clients.Group(tenantGroupId)
+                                            .SendAsync("ReceiveData", message, cancellationToken: stoppingToken)
+                                    );
                                 }
                                 else
                                 {
                                     // Chat messages (with text content)
                                     _logger.LogDebug("Sending message to group {GroupId}: {Message}",
                                         groupId, JsonSerializer.Serialize(message));
-                                    // TODO: Remove the backward compatibility ReceiveMessage later
-                                    await _hubContext.Clients.Group(groupId)
-                                        .SendAsync("ReceiveMessage", message, cancellationToken: stoppingToken);
-                                    // New method names
-                                    await _hubContext.Clients.Group(groupId)
-                                        .SendAsync("ReceiveChat", message, cancellationToken: stoppingToken);
-                                    await _tenantHubContext.Clients.Group(tenantGroupId)
-                                        .SendAsync("ReceiveChat", message, cancellationToken: stoppingToken);
+                                    await Task.WhenAll(
+                                        // TODO: Remove the backward compatibility ReceiveMessage later
+                                        _hubContext.Clients.Group(groupId)
+                                            .SendAsync("ReceiveMessage", message, cancellationToken: stoppingToken),
+                                        // New method names
+                                        _hubContext.Clients.Group(groupId)
+                                            .SendAsync("ReceiveChat", message, cancellationToken: stoppingToken),
+                                        _tenantHubContext.Clients.Group(tenantGroupId)
+                                            .SendAsync("ReceiveChat", message, cancellationToken: stoppingToken)
+                                    );
                                 }
 
                                 // Publish to SSE subscribers
@@ -213,7 +221,7 @@ namespace Features.UserApi.Services
                 catch (OperationCanceledException)
                 {
                     _logger.LogInformation("MongoChangeStreamService is stopping.");
-                    break; 
+                    break;
                 }
                 catch (MongoException ex) when (ex.HasErrorLabel("TransientTransactionError") || ex is MongoConnectionException || ex is MongoNotPrimaryException || ex is MongoNodeIsRecoveringException)
                 {
--- a/XiansAi.Server.Src/Features/UserApi/Utils/SSEEventWriter.cs
+++ b/XiansAi.Server.Src/Features/UserApi/Utils/SSEEventWriter.cs
@@ -8,6 +8,14 @@ namespace Features.UserApi.Utils;
 /// </summary>
 public static class SSEEventWriter
 {
+    // perf: reuse a single JsonSerializerOptions instance instead of allocating a new one
+    // on every SSE write. JsonSerializerOptions construction is expensive (~2-3 µs + GC
+    // pressure) and happens on every outgoing message on every connected SSE client.
+    private static readonly JsonSerializerOptions _jsonOptions = new(JsonSerializerDefaults.Web)
+    {
+        PropertyNamingPolicy = JsonNamingPolicy.CamelCase
+    };
+
     /// <summary>
     /// Writes a Server-Sent Event to the response stream
     /// </summary>
@@ -17,10 +25,7 @@ public static class SSEEventWriter
     /// <param name="cancellationToken">Cancellation token</param>
     public static async Task WriteEventAsync(HttpResponse response, string eventType, object data, CancellationToken cancellationToken)
     {
-        var json = JsonSerializer.Serialize(data, new JsonSerializerOptions
-        {
-            PropertyNamingPolicy = JsonNamingPolicy.CamelCase
-        });
+        var json = JsonSerializer.Serialize(data, _jsonOptions);
 
         var sseData = $"event: {eventType}\ndata: {json}\n\n";
         var bytes = Encoding.UTF8.GetBytes(sseData);
@@ -67,4 +72,4 @@ public static class SSEEventWriter
             subscriberCount
         };
     }
-} 
+}
--- a/XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs
+++ b/XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs
@@ -263,17 +263,20 @@ public class ConversationRepository : IConversationRepository
     private readonly ITenantContext _tenantContext;
     private readonly ISecureEncryptionService _encryptionService;
     private readonly string _uniqueSecret;
+    private readonly IBackgroundTaskService _backgroundTaskService;
 
     public ConversationRepository(
         IDatabaseService databaseService, 
         ILogger<ConversationRepository> logger, 
         ITenantContext tenantContext,
         ISecureEncryptionService encryptionService,
-        IConfiguration configuration)
+        IConfiguration configuration,
+        IBackgroundTaskService backgroundTaskService)
     {
         _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         _tenantContext = tenantContext ?? throw new ArgumentNullException(nameof(tenantContext));
         _encryptionService = encryptionService ?? throw new ArgumentNullException(nameof(encryptionService));
+        _backgroundTaskService = backgroundTaskService ?? throw new ArgumentNullException(nameof(backgroundTaskService));
         
         var database = databaseService.GetDatabaseAsync().GetAwaiter().GetResult();
         _database = database;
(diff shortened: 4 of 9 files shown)


hasith:
## Review in Progress

A comprehensive automated PR review has been started for this pull request. Specialized sub-agents are being coordinated to analyze:

- Code quality, readability, and maintainability
- Security vulnerabilities and input validation
- Test coverage and test quality
- Performance bottlenecks and resource usage

Results will be posted shortly.

hasith (commented):
## Automated PR Review — perf: Improving conversation message cycle time

**Verdict: APPROVED WITH CONDITIONS**
**Scope:** 9 files changed, 10 performance commits
**Risk Level:** MEDIUM-HIGH (touches encryption, DB transaction semantics, process-lifetime caches)

---

## Executive Summary

This PR delivers a well-structured set of 10 targeted performance optimizations addressing the bottlenecks catalogued in issue #396. The changes cover CPU (PBKDF2 caching), MongoDB I/O (transaction removal, query merging, new indexes), GC pressure (static JsonSerializerOptions), log overhead (conditional serialization), and latency (parallel SignalR sends). The engineering rationale for each change is sound and clearly documented.

However, four issues require attention before this is production-safe: one is a **security concern about plaintext key material in logs** that is pre-existing but made more visible by this PR, one is a **correctness risk in the validation failure cache**, one relates to **key rotation being silently blocked**, and one is a **process-lifetime cache with no invalidation path that could cause privilege escalation in edge cases**.

---

## CRITICAL Findings

### SEC-1: Derived AES-256 Key Material Held in Process Memory Indefinitely — No Key Rotation Path
**File:** `XiansAi.Server.Src/Shared/Services/SecureEncryptionService.cs`, line 29

```csharp
private static readonly ConcurrentDictionary<(string, string), byte[]> _derivedKeyCache = new();
```

The 32-byte derived AES key is stored as a plain `byte[]` in a static dictionary for the process lifetime with no zeroing, eviction, or TTL. Two concrete risks:

1. **Memory dump exposure:** If the process crashes or is inspected (e.g., by a debugger or production crash dump), the raw key material is immediately readable.
2. **Key rotation is blocked silently:** If `EncryptionKeys:BaseSecret` is rotated (e.g., after a suspected compromise), the old derived key remains in the static cache. New configuration is loaded but the cache still returns the pre-rotation key for any previously-seen `(baseSecret, uniqueSecret)` pair, **preventing the rotation from taking effect without a process restart.**

**Recommended fix:**
```csharp
// Option A: Document the restart requirement explicitly in code + ops runbook
// Add to startup validation:
_logger.LogWarning("NOTE: EncryptionKeys:BaseSecret changes require a process restart to take effect (derived key is cached process-lifetime).");

// Option B: Add a cache invalidation method and call it from a config-reload hook
public static void InvalidateDerivedKeyCache() => _derivedKeyCache.Clear();
```

At minimum, add an XML doc comment and a startup log warning so operators know a restart is required on key rotation.

---

### SEC-4: Validation Failure Results Are Cached in ActivationValidationService
**File:** `XiansAi.Server.Src/Shared/Services/ActivationValidationService.cs`, lines 85–91

```csharp
var workflowCheck = await _cache.GetOrAddAsync(
    workflowTypeCacheKey,
    _ => ValidateWorkflowTypeForAgentAsync(tenantId, agentName, workflowType.Trim()),
    WorkflowTypeCacheDuration,    // 5-minute TTL
    size: 1);
```

`GetOrAddAsync` caches whatever `ValidateWorkflowTypeForAgentAsync` returns — including `ServiceResult.Failure(...)`. This means:
- If a flow definition lookup fails transiently (network error returns an internal-server-error result), that failure is cached for 5 minutes.
- If an agent's flow definition is added/updated during the 5-minute TTL window, legitimate callers receive `BadRequest` until the cache expires.

**Standard mitigation:** Only cache `ServiceResult.Success()` results. Wrap the factory:

```csharp
var workflowCheck = await _cache.GetOrAddAsync(
    workflowTypeCacheKey,
    async _ => {
        var r = await ValidateWorkflowTypeForAgentAsync(tenantId, agentName, workflowType.Trim());
        return r.IsSuccess ? r : throw new InvalidOperationException("Not cacheable");
    },
    WorkflowTypeCacheDuration,
    size: 1);
// Fall back to direct call if cache throws
```

Or use a separate cache-aside pattern that only populates on success.

---

### SEC-6: Decrypted Conversation Text Logged at Information Level
**File:** `XiansAi.Server.Src/Features/UserApi/Services/MongoChangeStreamService.cs`, line 339

```csharp
_logger.LogInformation("Successfully decrypted message {MessageId}. Decrypted text: {Text}", message.Id, decryptedText);
```

Every decrypted conversation message is written to the application log at Information level. If logs are shipped to Elastic, Splunk, Datadog, or similar, this places conversation plaintext in a log store that almost certainly has weaker access controls and longer retention than the encrypted MongoDB collection. This defeats the purpose of AES-GCM encryption.

**Fix:** Downgrade to Debug level and omit the text value:
```csharp
_logger.LogDebug("Successfully decrypted message {MessageId}", message.Id);
```

This is pre-existing code but the PR touches this file and therefore review scope includes it.

---

## WARNING Findings

### SEC-3: `_systemAgentCache` Has No TTL or Invalidation — Stale System-Scope Routing Risk
**File:** `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs`, lines 76, 191–201

```csharp
private static readonly ConcurrentDictionary<string, bool> _systemAgentCache = new(StringComparer.Ordinal);
```

Agent system-scope is described as "immutable at runtime" in the commit message, but there is no enforcement of this invariant. If an agent is demoted from system-scope in the database, the cached `true` value means it continues receiving system-scoped Temporal workflow routing for the process lifetime, potentially bypassing tenant isolation.

**Recommended fix:** Add a 30-minute TTL or an explicit cache invalidation method called by the agent management API on scope changes.

---

### CQ-6: Widened DB Query Condition in MessageService.ProcessOutgoingMessage
**File:** `XiansAi.Server.Src/Shared/Services/MessageService.cs`, line 275

```csharp
// Old: only queried when Origin was missing
if (string.IsNullOrEmpty(request.Origin))

// New: queries when EITHER Origin is missing OR Data is null
if (string.IsNullOrEmpty(request.Origin) || request.Data == null)
```

The new condition fires `GetLastIncomingOriginAndDataAsync` (a MongoDB query) when `request.Data == null` even if `request.Origin` is already set to a non-app origin (e.g., `"web:..."`, `"email:..."`). The inner guard `request.Origin.StartsWith("app:")` prevents *assigning* `lastData`, but the query is still issued. In the original code, `GetLastIncomingDataAsync` was only called when `Origin` was already determined to start with `"app:"`.

This partially reverses the performance optimization it claims to make: web-origin messages now always pay the merged query cost even though they will never use `lastData`.

**Recommended fix:** Narrow the condition back:
```csharp
if (string.IsNullOrEmpty(request.Origin) || 
    (request.Data == null && !string.IsNullOrEmpty(request.Origin) && request.Origin.StartsWith("app:")))
```

Or restructure to keep the semantics of the original two-phase check while using the merged method.

---

### PERF-1: BackgroundTaskService Channel Is Unbounded — Heap Growth Under Burst Load
**File:** `XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs`, lines 473–478

Every message insert now enqueues a thread-timestamp update to `BackgroundTaskService`. If the background consumer serializes writes (SingleReader channel) and MongoDB is slow or recovering, the channel queue grows without bound. The PR description acknowledges this as a known follow-up but the risk is worth flagging explicitly: under a burst of 10,000 messages, this is 10,000 queued closures holding references to `IMongoCollection<ConversationThread>` and filter/update builders.

**Recommended:** Apply `BoundedChannel<T>` with capacity + `DropOldest` policy:
```csharp
var channel = Channel.CreateBounded<Func<Task>>(new BoundedChannelOptions(10_000)
{
    FullMode = BoundedChannelFullMode.DropOldest,
    SingleReader = true
});
```
Log a warning when an item is dropped.

---

### TEST-3: No Test Coverage for Removed MongoDB Transaction
**File:** `XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs`, lines 461–478

The transaction removal is the highest-impact behavioral change in this PR. Without integration tests, there is no automated verification that:
1. `thread.UpdatedAt` is eventually updated asynchronously
2. `MongoRetryHelper` retrying `InsertOneAsync` on a timeout does not produce a duplicate document (the old transaction guaranteed exactly-once insert; the new code does not)
3. The system behaves correctly when `BackgroundTaskService` drops the update under load

**Action required:** Add at minimum an integration test that verifies `thread.UpdatedAt` is updated within a reasonable time after message insert.

---

### TEST-2: PBKDF2 Key Cache Correctness Is Untested
**File:** `XiansAi.Server.Src/Shared/Services/SecureEncryptionService.cs`

The key caching change is the second-highest-impact change. A unit test verifying `Encrypt(x) |> Decrypt == x` with the cached path, and that two different `uniqueSecret` values produce distinct cache entries, would take 15 minutes to write and would prevent a regression if the cache key logic is ever changed.

---

## Suggestions (Non-Blocking)

### CQ-3: Use Idiomatic Pattern for Async Cache Population in WorkflowSignalService
**File:** `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs`, lines 191–200

```csharp
// Current: manual TryGetValue + assignment — benign race on first call per agent
if (_systemAgentCache.TryGetValue(agentName, out var cached))
    return cached;
var result = (await _agentService.IsSystemAgent(agentName)).Data;
_systemAgentCache[agentName] = result;
return result;
```

Under concurrent signals for a new agent name, multiple threads will simultaneously call `IsSystemAgent`. The results are identical (deterministic), so no correctness issue — but it's non-idiomatic. Consider `LazyInitializer` or a `SemaphoreSlim`-guarded pattern for clarity.

---

### CQ-7: No Invalidation Path for Workflow Type Cache
**File:** `XiansAi.Server.Src/Shared/Services/ActivationValidationService.cs`

`InvalidateActivationCache` invalidates the activation entity cache but not the new workflow-type cache. If an operator adds a new flow definition for an agent while the server is running, the `"not registered"` result (if cached) will persist for 5 minutes. Expose `InvalidateWorkflowTypeCache(tenantId, agentName, workflowType)` or call the workflow-type cache removal inside the existing `InvalidateActivationCache` method.

---

### PERF-6: Redundant CamelCase Assignment in SSEEventWriter
**File:** `XiansAi.Server.Src/Features/UserApi/Utils/SSEEventWriter.cs`, line 14

```csharp
private static readonly JsonSerializerOptions _jsonOptions = new(JsonSerializerDefaults.Web)
{
    PropertyNamingPolicy = JsonNamingPolicy.CamelCase  // redundant — Web defaults already set this
};
```

`JsonSerializerDefaults.Web` already applies `CamelCase` naming. The explicit assignment is harmless but misleads readers into thinking it is required.

---

### CQ-4: Indentation Defect in ConversationRepository.cs
**File:** `XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs`, after line 373

The `/// <summary>` XML doc comment on the `AddScopeFilter` helper method has extra leading indentation after `GetLastIncomingOriginAndDataAsync` is inserted above it (visible in diff line 375). This will cause `dotnet-format` or Roslyn analyzers to emit a warning.

---

## Confirmed Correct Optimizations

| Change | Assessment |
|---|---|
| PBKDF2 key caching via `ConcurrentDictionary.GetOrAdd` + static lambda | Correct. GetOrAdd factory double-invocation under contention is benign (deterministic result). |
| Static `JsonSerializerOptions` in `SSEEventWriter` | Correct. `JsonSerializerOptions` is thread-safe once sealed after first use. |
| `Task.WhenAll` for SignalR sends | Correct. Sends are independent; partial failure behavior is equivalent to or better than sequential. |
| `thread_chat_message_lookup` index `(tenant_id, thread_id, message_type, created_at DESC)` | Correct and optimal for the chatOnly=true query shape. |
| `outgoing_request_id_lookup` sparse index | Correct. Sparse index appropriate since most messages have no `request_id`. |
| `logger.IsEnabled(LogLevel.Information)` guard | Correct. Also a security improvement (avoids logging request bodies). |
| `_collectionEnsured` static volatile bool | Correct for singleton BackgroundService. Volatile read prevents stale cache-line reads. |
| Duplicate `background: true` key removal in `usage_metrics` | Correct fix for YAML authoring error. |

---

## Verification Checklist Additions

In addition to the items already listed in the PR body, please verify:

- [ ] Rotate `EncryptionKeys:BaseSecret` in a test environment and confirm new messages decrypt correctly (requires process restart — document this in runbook)
- [ ] Confirm `_logger.LogInformation("...Decrypted text: {Text}...")` in `MongoChangeStreamService` is downgraded to `LogDebug` before merging
- [ ] Verify `GetLastIncomingOriginAndDataAsync` query plan for `scope = null` case (explain plan shows index usage)
- [ ] Add `InvalidateWorkflowTypeCache` method or call cache removal from `InvalidateActivationCache`
- [ ] Confirm `MongoRetryHelper` does not retry on `MongoDuplicateKeyException`

---

*Review generated by automated PR review agent — 4 specialized sub-agents (code-reviewer, security-reviewer, test-reviewer, performance-reviewer) coordinated in parallel.*

hasith:
## Automated PR Review — perf: Improving conversation message cycle time

**Verdict: APPROVED WITH CONDITIONS**
**Scope:** 9 files changed, 10 performance commits
**Risk Level:** MEDIUM-HIGH (touches encryption, DB transaction semantics, process-lifetime caches)

---

## Executive Summary

This PR delivers a well-structured set of 10 targeted performance optimizations addressing the bottlenecks catalogued in issue #396. The changes cover CPU (PBKDF2 caching), MongoDB I/O (transaction removal, query merging, new indexes), GC pressure (static JsonSerializerOptions), log overhead (conditional serialization), and latency (parallel SignalR sends). The engineering rationale for each change is sound and clearly documented.

However, four issues require attention before this is production-safe: one is a **security concern about plaintext decrypted message text in logs**, one is a **correctness risk where validation failures are cached**, one relates to **key rotation being silently blocked by the process-lifetime cache**, and one is a **process-lifetime system-agent cache with no invalidation path** that could cause stale privilege routing in edge cases.

---

## CRITICAL Findings

### SEC-1: Derived AES-256 Key Material Held in Process Memory with No Rotation Path
**File:** `XiansAi.Server.Src/Shared/Services/SecureEncryptionService.cs`, line 29

```csharp
private static readonly ConcurrentDictionary<(string, string), byte[]> _derivedKeyCache = new();
```

The 32-byte derived AES key is stored as a plain `byte[]` in a static dictionary for the process lifetime with no zeroing, eviction, or TTL. Two concrete risks:

1. **Memory dump exposure:** If the process crashes or is inspected, the raw key bytes are immediately readable from the heap.
2. **Key rotation is silently blocked:** If `EncryptionKeys:BaseSecret` is rotated after a suspected compromise, the old derived key remains in the static cache for any previously-seen `(baseSecret, uniqueSecret)` pair. New configuration is loaded but `GetOrDeriveKey` returns the pre-rotation cached key — the rotation has no effect without a process restart.

**Recommended fix:** At minimum, add an XML doc comment and a startup log warning making the restart requirement explicit. Optionally expose a `InvalidateDerivedKeyCache()` method callable from a config-reload hook.

---

### SEC-4: Validation Failure Results Are Cached in ActivationValidationService
**File:** `XiansAi.Server.Src/Shared/Services/ActivationValidationService.cs`, lines 85–91

```csharp
var workflowCheck = await _cache.GetOrAddAsync(
    workflowTypeCacheKey,
    _ => ValidateWorkflowTypeForAgentAsync(tenantId, agentName, workflowType.Trim()),
    WorkflowTypeCacheDuration,    // 5-minute TTL
    size: 1);
```

`GetOrAddAsync` caches whatever the factory returns — including `ServiceResult.Failure(...)`. This means:
- If `ValidateWorkflowTypeForAgentAsync` returns a transient internal-server-error, that failure is cached for 5 minutes and all subsequent calls for the same `(tenantId, agentName, workflowType)` are denied without a DB lookup.
- If an agent's flow definition is added while the server is running, legitimate callers receive `BadRequest` until the 5-minute TTL expires.

**Standard mitigation:** Only cache success results. Failure results should always fall through to a fresh DB query.

---

### SEC-6: Decrypted Conversation Text Logged at Information Level
**File:** `XiansAi.Server.Src/Features/UserApi/Services/MongoChangeStreamService.cs`, line 339

```csharp
_logger.LogInformation("Successfully decrypted message {MessageId}. Decrypted text: {Text}", message.Id, decryptedText);
```

Every decrypted conversation message is written to application logs at Information level in plaintext. If logs are shipped to Elastic, Splunk, Datadog, or similar aggregators, conversation content is stored there in cleartext, defeating the purpose of AES-GCM encryption at rest.

**Fix:** Downgrade to Debug level and omit the plaintext value:
```csharp
_logger.LogDebug("Successfully decrypted message {MessageId}", message.Id);
```

This is pre-existing code but this PR touches `MongoChangeStreamService.cs` and therefore falls within review scope.

---

## WARNING Findings

### SEC-3: `_systemAgentCache` Has No TTL or Invalidation — Stale Routing Risk
**File:** `XiansAi.Server.Src/Shared/Services/WorkflowSignalService.cs`, lines 76, 191–201

The commit message states agent system-scope is "immutable at runtime" but there is no database-level enforcement. If an agent is demoted from system-scope, the cached `true` result means it continues receiving system-scoped Temporal workflow routing for the rest of the process lifetime, potentially bypassing tenant isolation.

**Recommended:** Add a 30-minute TTL or an explicit `InvalidateSystemAgentCache(agentName)` method called by the agent management API when scope attributes change.

---

### CQ-6: Widened DB Query Condition Partially Reverses the Optimization
**File:** `XiansAi.Server.Src/Shared/Services/MessageService.cs`, line 275

```csharp
// Old: only queried when Origin was missing
if (string.IsNullOrEmpty(request.Origin))

// New: queries when EITHER Origin is missing OR Data is null
if (string.IsNullOrEmpty(request.Origin) || request.Data == null)
```

The new condition fires `GetLastIncomingOriginAndDataAsync` (a MongoDB query) when `request.Data == null` even if `request.Origin` is already set to a non-app origin (e.g., `"web:..."`, `"email:..."`). The inner guard `request.Origin.StartsWith("app:")` prevents *assigning* `lastData` in that case, but the query is still issued. Web-origin messages now always pay the merged query cost even though `lastData` is never used.

**Recommended fix:** Narrow the outer condition or keep a two-phase check — only call the merged method when there is actually something to look up.

---

### PERF-1: BackgroundTaskService Channel Is Unbounded — Heap Growth Under Burst Load
**File:** `XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs`, lines 473–478

Every message insert enqueues a thread-timestamp update to `BackgroundTaskService`. If the background consumer falls behind (MongoDB slow/recovering), the channel queue grows without bound — each queued item holds a closure over `IMongoCollection<ConversationThread>` and filter/update builder objects. Under a burst of 10,000 messages this is 10,000 queued closures accumulating in heap.

**Recommended:** Apply `BoundedChannel<T>` with a reasonable capacity and `DropOldest` or `Wait` policy. Log a warning when an item is dropped.

---

### TEST-3: No Test Coverage for Removed MongoDB Transaction (Highest-Impact Behavioral Change)
**File:** `XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs`, lines 461–478

The transaction removal is the highest-impact behavioral change in the PR. Without integration tests there is no automated verification that:
1. `thread.UpdatedAt` is eventually updated asynchronously after message insert
2. `MongoRetryHelper` retrying `InsertOneAsync` on a network timeout does not produce a duplicate document (the old transaction guaranteed exactly-once insert; the new code does not)
3. The system behaves correctly when `BackgroundTaskService` drops the update under load

**Action required:** Add at minimum one integration test verifying `thread.UpdatedAt` is updated within a reasonable window after `SaveMessageAsync` returns.

---

### TEST-2: PBKDF2 Key Cache Correctness Is Untested
**File:** `XiansAi.Server.Src/Shared/Services/SecureEncryptionService.cs`

The key caching change is the second-highest-impact change in the PR. A unit test verifying `Decrypt(Encrypt(x)) == x` using the cached path, and that two different `uniqueSecret` values produce distinct cached keys (and thus distinct ciphertext), would prevent a regression if the cache key logic changes.

---

## Suggestions (Non-Blocking)

### CQ-7: No Invalidation Path for Workflow Type Cache
`InvalidateActivationCache` invalidates the activation entity cache but not the new workflow-type cache. If a flow definition is added while the server is running, `"not registered"` errors persist for 5 minutes. Expose `InvalidateWorkflowTypeCache(tenantId, agentName, workflowType)` or call it from inside `InvalidateActivationCache`.

### CQ-3: Idiomatic Pattern for Async Cache Population in WorkflowSignalService
The manual `TryGetValue` + assignment pattern at lines 192–200 has a benign race (both threads compute the same deterministic result). Consider `LazyInitializer` or a `SemaphoreSlim`-guarded pattern for clarity and to avoid the double-call on first access.

### PERF-6: Redundant CamelCase Assignment in SSEEventWriter
`JsonSerializerDefaults.Web` already applies `CamelCase` naming. The explicit `PropertyNamingPolicy = JsonNamingPolicy.CamelCase` assignment on line 16 is redundant and may mislead reviewers into thinking it is required.

### CQ-4: Indentation Defect in ConversationRepository.cs
The `/// <summary>` XML doc comment on `AddScopeFilter` (line ~375) has extra leading indentation introduced when `GetLastIncomingOriginAndDataAsync` was inserted above it. Will fail `dotnet-format` / Roslyn style analyzer.

---

## Confirmed Correct Optimizations

| Change | Assessment |
|---|---|
| PBKDF2 key caching via `ConcurrentDictionary.GetOrAdd` + static lambda | Correct. Double-invocation under contention is benign (deterministic KDF result). |
| Static `JsonSerializerOptions` in `SSEEventWriter` | Correct. Thread-safe once sealed after first serialization call. |
| `Task.WhenAll` for SignalR sends | Correct. Sends are independent; partial failure behavior is equivalent to or better than sequential. |
| `thread_chat_message_lookup` index `(tenant_id, thread_id, message_type, created_at DESC)` | Correct and optimal for the chatOnly=true query shape. |
| `outgoing_request_id_lookup` sparse index | Correct. Sparse index appropriate since most messages have no `request_id`. |
| `logger.IsEnabled(LogLevel.Information)` guard in MessagingEndpoints | Correct. Also an incidental security improvement (avoids logging full request body). |
| `_collectionEnsured` static volatile bool | Correct for singleton BackgroundService. `volatile` ensures cross-thread visibility without a lock. |
| Duplicate `background: true` key removal in `usage_metrics` YAML | Correct — was a YAML authoring error with no functional effect. |

---

## Additional Verification Checklist Items

Beyond the items in the PR body, please verify:

- [ ] Rotate `EncryptionKeys:BaseSecret` in a test environment and confirm new messages can no longer be decrypted with the old key after rotation (i.e., that a process restart is required — and document this in the ops runbook)
- [ ] Downgrade `LogInformation("...Decrypted text: {Text}...")` in `MongoChangeStreamService` to `LogDebug` before merging
- [ ] Run `EXPLAIN` on `GetLastIncomingOriginAndDataAsync` with `scope = null` to confirm index coverage
- [ ] Add `InvalidateWorkflowTypeCache` method or integrate into `InvalidateActivationCache`
- [ ] Confirm `MongoRetryHelper` does not retry on `MongoDuplicateKeyException`

---

*Review generated by automated PR review agent — 4 specialized sub-agents (code-reviewer, security-reviewer, test-reviewer, performance-reviewer) coordinated in parallel. Base: `1b0fa6b` — Head: `2430bcd`.*

PR #401, opened 2026-05-08
Implement processed event tracking for conversation messages in Mongo…

Diyanada99x:
…ChangeStreamService

### DIFF
--- a/XiansAi.Server.Src/Features/UserApi/Services/MongoChangeStreamService.cs
+++ b/XiansAi.Server.Src/Features/UserApi/Services/MongoChangeStreamService.cs
@@ -43,7 +43,7 @@ namespace Features.UserApi.Services
             _tenantHubContext = tenantHubContext;
             _messageEventPublisher = messageEventPublisher ?? throw new ArgumentNullException(nameof(messageEventPublisher));
             _encryptionService = encryptionService ?? throw new ArgumentNullException(nameof(encryptionService));
-
+        
             // Get the unique secret for conversation messages
             _uniqueSecret = configuration["EncryptionKeys:UniqueSecrets:ConversationMessageKey"] ?? string.Empty;
             if (string.IsNullOrWhiteSpace(_uniqueSecret))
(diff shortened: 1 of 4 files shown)


XiansAi.Server.Src/Features/UserApi/Services/MongoChangeStreamService.cs
```diff
@@ -133,6 +133,19 @@ await MongoRetryHelper.ExecuteWithRetryAsync(async () =>
                             var message = changeDoc.FullDocument;
                             if (message == null) continue;
 
+                            if (message.Scope == "Slack" || message.Scope == "Teams")
```
hasith: I think a better way to identify is to use 
request.Origin.StartsWith("app:")

See: XiansAi.Server.Src/Shared/Services/MessageService.cs

XiansAi.Server.Src/Features/UserApi/Services/MongoChangeStreamService.cs
```diff
@@ -67,7 +67,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                     using var scope = _scopeFactory.CreateScope();
                     var databaseService = scope.ServiceProvider.GetRequiredService<IDatabaseService>();
                     var pendingRequestService = scope.ServiceProvider.GetRequiredService<IPendingRequestService>();
-
+                    var processedEventRepository = scope.ServiceProvider.GetRequiredService<IProcessedEventRepository>();
```
hasith: If 'processedEventRepository' is used only in the if condition below, move this statement into that.

hasith (changes requested):

XiansAi.Server.Src/Features/UserApi/Services/MongoChangeStreamService.cs
Diyanada99x: done. since the latest deployment working fine without these changes im not going to deploy a new one to prod. locally tested.

PR #415, opened 2026-06-15
Security/fix redact email domain sanitize

hasith:


hasith (commented):
## 🔍 Automated PR Review — PR #415

**Verdict: ✅ APPROVE**

---

### Summary

This is a focused, well-crafted security hardening PR comprising three independent fixes:

1. **Log-injection fix in `RedactEmail`** — adds CR/LF stripping to the domain portion of a redacted email address, closing a genuine log-forging gap (CWE-117) that existed in the domain segment while the local-part was already masked.
2. **CodeQL neutralModel schema fix** — removes an extraneous `subtypes` boolean column that was causing every row to be silently discarded, meaning the CodeQL sanitizer models were doing nothing until this fix.
3. **CI improvements** — upgrades `codeql-action` v3→v4 (v3 is deprecated) and adds `workflow_dispatch` for manual triggering.

All three changes are correct and intentional. No regressions or security vulnerabilities were introduced.

---

### 🟢 Critical Issues
None.

### 🟡 Warnings
None.

### 💡 Suggestions

#### 1. Missing regression test for the `RedactEmail` log-injection fix
**File:** `XiansAi.Server.Src/Shared/Utils/LogSanitizer.cs` — line 41

`RedactEmail` was silently vulnerable to log injection via the domain portion, and no unit tests for this method currently exist in the test suite. The fix is correct, but a regression test would document the security requirement and guard against it being accidentally re-introduced.

Example test case:
```csharp
[Fact]
public void RedactEmail_StripsCrLfFromDomain()
{
    // Arrange: attacker-controlled domain with embedded newlines
    var malicious = "user@evil.com\r\nFAKE_LOG_ENTRY";

    // Act
    var result = LogSanitizer.RedactEmail(malicious);

    // Assert: newlines replaced with spaces, no fake log lines possible
    Assert.Equal("***@evil.com  FAKE_LOG_ENTRY", result);
    Assert.DoesNotContain("\r", result);
    Assert.DoesNotContain("\n", result);
}
```

---

### ✅ Positive Notes

- The CR/LF replacement in `RedactEmail` exactly mirrors the pattern used in `Sanitize()` — consistent approach throughout the class.
- The CodeQL model column comment added in `LogSanitizerModels.yml` (`# neutralModel columns: namespace, type, name, signature, kind, provenance`) is a good self-documenting addition that will prevent this mistake from recurring.
- Upgrading to `codeql-action@v4` is the right call — v3 is actively being deprecated by GitHub.
- The `extensionTargets` / `dataExtensions` additions to `codeql-pack.yml` are required for the model pack to be discovered by CodeQL, so these were prerequisite to the model fix working at all.
- Removing the leading `./` from the `model-packs` path in `codeql-config.yml` is a minor but correct normalization.

---

*🤖 Reviewed by [Claude Code](https://claude.com/claude-code) automated PR review*

XiansAi.Server.Src/Shared/Utils/LogSanitizer.cs
```diff
@@ -38,6 +38,7 @@ public static string RedactEmail(string? email)
         if (atIndex < 0)
             return "***@[no-domain]";
 
-        return "***@" + email[(atIndex + 1)..];
+        var domain = email[(atIndex + 1)..].Replace('\r', ' ').Replace('\n', ' ');
```
hasith, line 41: **💡 Suggestion — Missing regression test for this security fix**

This method now correctly strips CR/LF from the domain, closing a genuine log-injection vector (CWE-117). However, no unit test exists for `RedactEmail` anywhere in the test suite.

A regression test would prevent this from being accidentally re-introduced and documents the security invariant explicitly. Example:

```csharp
[Fact]
public void RedactEmail_StripsCrLfFromDomain()
{
    var malicious = "user@evil.com\r\nFAKE_LOG_ENTRY";
    var result = LogSanitizer.RedactEmail(malicious);
    Assert.DoesNotContain("\r", result);
    Assert.DoesNotContain("\n", result);
}
```

PR #425, opened 2026-07-01
Enhance webhook functionality and configuration

hasith:
- Added outbound webhook options to the .env.example file, allowing for event delivery to configured listener URLs.
- Introduced new MongoDB indexes for managing webhook deliveries, including atomic claim queries and TTL cleanup.
- Integrated webhook event publishing in various services, including agent ownership transfer, flow definition updates, and activation management.
- Registered webhook services in SharedConfiguration to support webhook delivery infrastructure.

These changes improve the system's ability to handle and manage webhook events, enhancing overall functionality and reliability.

### DIFF
--- a/XiansAi.Server.Src/.env.example
+++ b/XiansAi.Server.Src/.env.example
@@ -129,6 +129,31 @@ Logging__LogLevel__Microsoft.AspNetCore=Warning
 UsageEvents__RecordUsageEvents=true
 
 # ====================
+# OUTBOUND WEBHOOKS (Optional)
+# ====================
+# Deliver server events (e.g. user.created, user.tenant.added, tenant.created,
+# user.role.changed) to configured listener URLs. Events are persisted to a Mongo outbox and
+# delivered by a background dispatcher that atomically claims each delivery, so exactly one
+# instance sends it across a multi-server deployment. See docs/OUTBOUND_WEBHOOKS.md.
+Webhooks__Enabled=false
+#Webhooks__PollIntervalSeconds=5
+#Webhooks__RequestTimeoutSeconds=10
+#Webhooks__MaxAttempts=5
+#Webhooks__LeaseSeconds=60
+#Webhooks__BatchSize=20
+# Require HTTPS listener URLs (recommended). Set true only for trusted internal http listeners.
+#Webhooks__AllowInsecureHttp=false
+# One entry per listener. EventTypes empty or "*" means all events.
+#Webhooks__Subscriptions__0__Name=crm-listener
+#Webhooks__Subscriptions__0__Url=https://example.com/hooks/xians
+#Webhooks__Subscriptions__0__Secret=shared-secret-for-hmac
+#Webhooks__Subscriptions__0__Enabled=true
+#Webhooks__Subscriptions__0__EventTypes__0=user.created
+#Webhooks__Subscriptions__0__EventTypes__1=user.tenant.added
+#Webhooks__Subscriptions__0__EventTypes__2=tenant.created
+#Webhooks__Subscriptions__0__EventTypes__3=user.role.changed
+
+# ====================
 # OPENTELEMETRY (Optional)
 # ====================
 # Set Enabled=true and provide an OtlpEndpoint to activate telemetry export.
--- a/XiansAi.Server.Src/Features/AgentApi/Services/DefinitionsService.cs
+++ b/XiansAi.Server.Src/Features/AgentApi/Services/DefinitionsService.cs
@@ -138,6 +138,7 @@ public class DefinitionsService : IDefinitionsService
     private readonly ITenantContext _tenantContext;
     private readonly IMarkdownService _markdownService; 
     private readonly IAgentPermissionRepository _agentPermissionRepository;
+    private readonly IWebhookEventPublisher _webhookEventPublisher;
     
     public DefinitionsService(
         Repositories.IFlowDefinitionRepository flowDefinitionRepository,
@@ -145,7 +146,8 @@ public class DefinitionsService : IDefinitionsService
         ILogger<DefinitionsService> logger,
         ITenantContext tenantContext,
         IMarkdownService markdownService,
-        IAgentPermissionRepository agentPermissionRepository
+        IAgentPermissionRepository agentPermissionRepository,
+        IWebhookEventPublisher webhookEventPublisher
     )
     {
         _flowDefinitionRepository = flowDefinitionRepository;
@@ -154,6 +156,7 @@ public class DefinitionsService : IDefinitionsService
         _tenantContext = tenantContext;
         _markdownService = markdownService;
         _agentPermissionRepository = agentPermissionRepository;
+        _webhookEventPublisher = webhookEventPublisher;
     }
 
     public async Task<IResult> CreateAsync(FlowDefinitionRequest request)
@@ -232,6 +235,12 @@ public class DefinitionsService : IDefinitionsService
                 // Create new definition with fresh ID
                 definition.Id = ObjectId.GenerateNewId().ToString();
                 await _flowDefinitionRepository.CreateAsync(definition);
+
+                await _webhookEventPublisher.PublishAsync(
+                    WebhookEventTypes.FlowDefinitionUpdated,
+                    new { tenantId = _tenantContext.TenantId, agentName = request.Agent, workflowType = definition.WorkflowType, systemScoped = request.SystemScoped, hash = definition.Hash },
+                    _tenantContext.TenantId);
+
                 return Results.Ok("Definition deleted and recreated successfully");
             }
             
@@ -241,6 +250,12 @@ public class DefinitionsService : IDefinitionsService
         _logger.LogInformation("Creating new definition {WorkflowType}", LogSanitizer.Sanitize(definition.WorkflowType));
         await GenerateMarkdown(definition);
         await _flowDefinitionRepository.CreateAsync(definition);
+
+        await _webhookEventPublisher.PublishAsync(
+            WebhookEventTypes.FlowDefinitionCreated,
+            new { tenantId = _tenantContext.TenantId, agentName = request.Agent, workflowType = definition.WorkflowType, systemScoped = request.SystemScoped, hash = definition.Hash },
+            _tenantContext.TenantId);
+
         return Results.Ok("New definition created successfully");
     }
 
@@ -295,6 +310,11 @@ public class DefinitionsService : IDefinitionsService
                 statusCode: StatusCodes.Status409Conflict);
         }
 
+        // Determine whether this is a brand-new agent (so we only publish agent.registered once,
+        // rather than on every worker restart which re-upserts the same agent).
+        var existingAgent = await _agentRepository.GetByNameInternalAsync(
+            request.AgentName, request.SystemScoped ? null! : _tenantContext.TenantId);
+
         // Create or update agent using thread-safe upsert operation
         var agent = await _agentRepository.UpsertAgentAsync(
             request.AgentName, 
@@ -310,6 +330,14 @@ public class DefinitionsService : IDefinitionsService
             request.SamplePrompts);
         
         _logger.LogInformation("Agent {AgentName} created/updated successfully by user {UserId}", LogSanitizer.Sanitize(request.AgentName), LogSanitizer.Sanitize(currentUser));
+
+        if (existingAgent == null)
+        {
+            await _webhookEventPublisher.PublishAsync(
+                WebhookEventTypes.AgentRegistered,
+                new { tenantId = _tenantContext.TenantId, agentId = agent.Id, agentName = agent.Name, systemScoped = agent.SystemScoped, createdBy = agent.CreatedBy },
+                _tenantContext.TenantId);
+        }
         
         return Results.Ok(new 
         { 
--- a/XiansAi.Server.Src/Shared/Configuration/Options/WebhooksOptions.cs
+++ b/XiansAi.Server.Src/Shared/Configuration/Options/WebhooksOptions.cs
@@ -1,0 +1,105 @@
+namespace Shared.Configuration;
+
+/// <summary>
+/// Configuration for the outbound webhook system. Bound from the "Webhooks" configuration section.
+/// Subscriptions (listener URLs) are defined entirely in configuration; secrets are never persisted.
+/// </summary>
+public class WebhooksOptions
+{
+    public const string SectionName = "Webhooks";
+
+    /// <summary>
+    /// Master switch. When false, no events are queued and the dispatcher does not run.
+    /// </summary>
+    public bool Enabled { get; set; } = false;
+
+    /// <summary>
+    /// How often the background dispatcher polls the outbox for due deliveries.
+    /// </summary>
+    public int PollIntervalSeconds { get; set; } = 5;
+
+    /// <summary>
+    /// Per-request HTTP timeout when POSTing to a listener URL.
+    /// </summary>
+    public int RequestTimeoutSeconds { get; set; } = 10;
+
+    /// <summary>
+    /// Maximum number of delivery attempts before a delivery is marked as failed.
+    /// </summary>
+    public int MaxAttempts { get; set; } = 5;
+
+    /// <summary>
+    /// How long a claimed delivery is leased to a single instance. If the instance crashes
+    /// before finishing, the lease expires and another instance can reclaim the delivery.
+    /// </summary>
+    public int LeaseSeconds { get; set; } = 60;
+
+    /// <summary>
+    /// Maximum number of deliveries claimed and processed per poll iteration.
+    /// </summary>
+    public int BatchSize { get; set; } = 20;
+
+    /// <summary>
+    /// When false (default), listener URLs must use HTTPS. Set to true only for trusted
+    /// internal listeners reachable over plain HTTP. Delivering event payloads (which may
+    /// contain PII) over HTTP exposes them to network eavesdropping.
+    /// </summary>
+    public bool AllowInsecureHttp { get; set; } = false;
+
+    /// <summary>
+    /// Configured webhook listener subscriptions.
+    /// </summary>
+    public List<WebhookSubscriptionOptions> Subscriptions { get; set; } = new();
+}
+
+/// <summary>
+/// A single configured webhook listener.
+/// </summary>
+public class WebhookSubscriptionOptions
+{
+    /// <summary>
+    /// Stable identifier for this subscription. Persisted on delivery rows so the dispatcher
+    /// can resolve the URL and secret from configuration at delivery time.
+    /// </summary>
+    public string Name { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Destination URL that receives the HTTP POST for each matching event.
+    /// </summary>
+    public string Url { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Optional shared secret. When set, an HMAC-SHA256 signature of the request body is sent
+    /// in the <c>X-Xians-Signature</c> header so the listener can verify authenticity.
+    /// </summary>
+    public string? Secret { get; set; }
+
+    /// <summary>
+    /// Event types this subscription receives. Empty or a single "*" entry means all events.
+    /// </summary>
+    public List<string> EventTypes { get; set; } = new();
+
+    /// <summary>
+    /// Whether this subscription is active.
+    /// </summary>
+    public bool Enabled { get; set; } = true;
+
+    /// <summary>
+    /// Returns true when this subscription should receive the given event type.
+    /// </summary>
+    public bool MatchesEvent(string eventType)
+    {
+        if (!Enabled)
+        {
+            return false;
+        }
+
+        if (EventTypes == null || EventTypes.Count == 0)
+        {
+            return true;
+        }
+
+        return EventTypes.Any(e =>
+            e == "*" || string.Equals(e, eventType, StringComparison.OrdinalIgnoreCase));
+    }
+}
--- a/XiansAi.Server.Src/Shared/Data/Models/WebhookDelivery.cs
+++ b/XiansAi.Server.Src/Shared/Data/Models/WebhookDelivery.cs
@@ -1,0 +1,318 @@
+using System.Text.Json.Serialization;
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+
+namespace Shared.Data.Models;
+
+/// <summary>
+/// Canonical event type identifiers for outbound webhooks. This is the catalog of all trackable
+/// events. Adding a new event is as simple as declaring a constant here and calling the publisher
+/// from the relevant service.
+///
+/// High-volume telemetry (per-message conversation writes, agent logs, activity history, usage
+/// metrics, ephemeral cache writes, heartbeats) is intentionally NOT tracked here, as it would
+/// overwhelm webhook listeners.
+/// </summary>
+public static class WebhookEventTypes
+{
+    // ----- Tenant lifecycle -----
+
+    /// <summary>A new tenant was created.</summary>
+    public const string TenantCreated = "tenant.created";
+
+    /// <summary>A tenant's profile (name/domain/description/theme/logo/timezone) was updated.</summary>
+    public const string TenantUpdated = "tenant.updated";
+
+    /// <summary>A tenant was enabled.</summary>
+    public const string TenantEnabled = "tenant.enabled";
+
+    /// <summary>A tenant was disabled.</summary>
+    public const string TenantDisabled = "tenant.disabled";
+
+    /// <summary>A tenant was deleted.</summary>
+    public const string TenantDeleted = "tenant.deleted";
+
+    /// <summary>A tenant's OIDC configuration was created or updated.</summary>
+    public const string TenantOidcUpdated = "tenant.oidc.updated";
+
+    /// <summary>A tenant's OIDC configuration was deleted.</summary>
+    public const string TenantOidcDeleted = "tenant.oidc.deleted";
+
+    // ----- User lifecycle (tenant-scoped and global) -----
+
+    /// <summary>A brand-new user account was created.</summary>
+    public const string UserCreated = "user.created";
+
+    /// <summary>An existing user account was granted membership in a tenant.</summary>
+    public const string UserTenantAdded = "user.tenant.added";
+
+    /// <summary>A user's membership in a tenant was removed.</summary>
+    public const string UserTenantRemoved = "user.tenant.removed";
+
+    /// <summary>A user's profile (name/email) was updated.</summary>
+    public const string UserUpdated = "user.updated";
+
+    /// <summary>A user's tenant membership was approved.</summary>
+    public const string UserApproved = "user.approved";
+
+    /// <summary>A user's tenant membership approval was revoked.</summary>
+    public const string UserUnapproved = "user.unapproved";
+
+    /// <summary>A role was added to a user within a tenant.</summary>
+    public const string UserRoleChanged = "user.role.changed";
+
+    /// <summary>A role was removed from a user within a tenant.</summary>
+    public const string UserRoleRemoved = "user.role.removed";
+
+    /// <summary>A user was granted the system administrator flag.</summary>
+    public const string UserSysAdminGranted = "user.sysadmin.granted";
+
+    /// <summary>A user's system administrator flag was revoked.</summary>
+    public const string UserSysAdminRevoked = "user.sysadmin.revoked";
+
+    /// <summary>A user account was enabled (unlocked).</summary>
+    public const string UserEnabled = "user.enabled";
+
+    /// <summary>A user account was disabled (locked out).</summary>
+    public const string UserDisabled = "user.disabled";
+
+    // ----- Agents, deployments and templates -----
+
+    /// <summary>An agent was registered for the first time (via the Agent API).</summary>
+    public const string AgentRegistered = "agent.registered";
+
+    /// <summary>An agent and its dependent resources were deleted.</summary>
+    public const string AgentDeleted = "agent.deleted";
+
+    /// <summary>An agent deployment's configuration was updated (via the Admin API).</summary>
+    public const string AgentDeploymentUpdated = "agent.deployment.updated";
+
+    /// <summary>Ownership of an agent was transferred to another user.</summary>
+    public const string AgentOwnershipTransferred = "agent.ownership.transferred";
+
+    /// <summary>A system template agent was deployed into a tenant.</summary>
+    public const string AgentTemplateDeployed = "agent.template.deployed";
+
+    /// <summary>A system-scoped template agent's metadata was updated.</summary>
+    public const string TemplateUpdated = "template.updated";
+
+    /// <summary>A system-scoped template agent was deleted.</summary>
+    public const string TemplateDeleted = "template.deleted";
+
+    // ----- Flow definitions -----
+
+    /// <summary>A new flow (workflow) definition was registered.</summary>
+    public const string FlowDefinitionCreated = "flow.definition.created";
+
+    /// <summary>An existing flow (workflow) definition was updated (hash changed).</summary>
+    public const string FlowDefinitionUpdated = "flow.definition.updated";
+
+    // ----- Activations -----
+
+    /// <summary>An agent activation was created.</summary>
+    public const string ActivationCreated = "activation.created";
+
+    /// <summary>An agent activation was updated.</summary>
+    public const string ActivationUpdated = "activation.updated";
+
+    /// <summary>An agent activation was activated (workflows started).</summary>
+    public const string ActivationActivated = "activation.activated";
+
+    /// <summary>An agent activation was deactivated.</summary>
+    public const string ActivationDeactivated = "activation.deactivated";
+
+    /// <summary>An agent activation was deleted.</summary>
+    public const string ActivationDeleted = "activation.deleted";
+
+    // ----- Knowledge -----
+
+    /// <summary>A knowledge item (or override/version) was created.</summary>
+    public const string KnowledgeCreated = "knowledge.created";
+
+    /// <summary>A knowledge item was updated (new version).</summary>
+    public const string KnowledgeUpdated = "knowledge.updated";
+
+    /// <summary>A knowledge item was deleted.</summary>
+    public const string KnowledgeDeleted = "knowledge.deleted";
+
+    // ----- Secrets -----
+
+    /// <summary>A vault secret was created.</summary>
+    public const string SecretCreated = "secret.created";
+
+    /// <summary>A vault secret was updated.</summary>
+    public const string SecretUpdated = "secret.updated";
+
+    /// <summary>A vault secret was deleted.</summary>
+    public const string SecretDeleted = "secret.deleted";
+
+    // ----- API keys and certificates -----
+
+    /// <summary>An API key was created.</summary>
+    public const string ApiKeyCreated = "apikey.created";
+
+    /// <summary>An API key was revoked.</summary>
+    public const string ApiKeyRevoked = "apikey.revoked";
+
+    /// <summary>An API key was rotated.</summary>
+    public const string ApiKeyRotated = "apikey.rotated";
+
+    /// <summary>A client certificate was issued.</summary>
+    public const string CertificateCreated = "certificate.created";
+
+    /// <summary>A client certificate was revoked.</summary>
+    public const string CertificateRevoked = "certificate.revoked";
+
+    // ----- App integrations -----
+
+    /// <summary>An app integration was created.</summary>
+    public const string IntegrationCreated = "integration.created";
+
+    /// <summary>An app integration was updated.</summary>
+    public const string IntegrationUpdated = "integration.updated";
+
+    /// <summary>An app integration was deleted.</summary>
+    public const string IntegrationDeleted = "integration.deleted";
+
+    /// <summary>An app integration was enabled.</summary>
+    public const string IntegrationEnabled = "integration.enabled";
+
+    /// <summary>An app integration was disabled.</summary>
+    public const string IntegrationDisabled = "integration.disabled";
+
+    /// <summary>A builtin webhook integration was created. (Deletion emits the generic integration.deleted event.)</summary>
+    public const string IntegrationWebhookCreated = "integration.webhook.created";
+}
+
+/// <summary>
+/// Lifecycle state of a single webhook delivery row.
+/// </summary>
+public enum WebhookDeliveryStatus
+{
+    /// <summary>Waiting to be delivered (or waiting for the next retry).</summary>
+    Pending,
+
+    /// <summary>Claimed by an instance and currently being delivered.</summary>
+    Delivering,
+
+    /// <summary>Successfully delivered (listener returned a 2xx response).</summary>
+    Delivered,
+
+    /// <summary>Permanently failed after exhausting the maximum number of attempts.</summary>
+    Failed
+}
+
+/// <summary>
+/// Identifies the authenticated principal that triggered an event, for auditing.
+/// </summary>
+public class WebhookActor
+{
+    /// <summary>The acting user's id (from the authenticated context), when known.</summary>
+    [JsonPropertyName("userId")]
+    public string? UserId { get; set; }
+
+    /// <summary>How the actor authenticated (e.g. UserToken, UserApiKey, AgentApiKey).</summary>
+    [JsonPropertyName("userType")]
+    public string? UserType { get; set; }
+
+    /// <summary>The tenant the actor was operating in (may differ from the event's target tenant).</summary>
+    [JsonPropertyName("tenantId")]
+    public string? TenantId { get; set; }
+
+    /// <summary>The actor's roles at the time of the action.</summary>
+    [JsonPropertyName("roles")]
+    public string[]? Roles { get; set; }
+}
+
+/// <summary>
+/// The JSON body POSTed to a webhook listener. <see cref="Data"/> carries event-specific fields.
+/// </summary>
+public class WebhookEventEnvelope
+{
+    [JsonPropertyName("eventType")]
+    public required string EventType { get; set; }
+
+    /// <summary>Unique id of the logical event; identical across all subscription deliveries.</summary>
+    [JsonPropertyName("eventId")]
+    public required string EventId { get; set; }
+
+    [JsonPropertyName("tenantId")]
+    public string? TenantId { get; set; }
+
+    [JsonPropertyName("occurredAt")]
+    public DateTime OccurredAt { get; set; }
+
+    /// <summary>Who triggered the event (for auditing). Null for system-originated events.</summary>
+    [JsonPropertyName("actor")]
+    public WebhookActor? Actor { get; set; }
+
+    [JsonPropertyName("data")]
+    public object? Data { get; set; }
+}
+
+/// <summary>
+/// Outbox row representing the delivery of one event to one subscription, persisted in the
+/// MongoDB collection <c>webhook_deliveries</c>. A background dispatcher atomically claims due
+/// rows so exactly one server instance delivers each one, then retries on failure.
+/// </summary>
+[BsonIgnoreExtraElements]
+public class WebhookDelivery
+{
+    [BsonId]
+    [BsonRepresentation(BsonType.ObjectId)]
+    public string Id { get; set; } = string.Empty;
+
+    /// <summary>Logical event id, shared across all subscription deliveries for the same event.</summary>
+    [BsonElement("event_id")]
+    public required string EventId { get; set; }
+
+    [BsonElement("event_type")]
+    public required string EventType { get; set; }
+
+    /// <summary>Name of the configured subscription this delivery targets.</summary>
+    [BsonElement("subscription_name")]
+    public required string SubscriptionName { get; set; }
+
+    [BsonElement("tenant_id")]
+    public string? TenantId { get; set; }
+
+    /// <summary>Id of the user who triggered the event (for auditing). Null for system events.</summary>
+    [BsonElement("actor_user_id")]
+    public string? ActorUserId { get; set; }
+
+    /// <summary>How the acting user authenticated (for auditing).</summary>
+    [BsonElement("actor_user_type")]
+    public string? ActorUserType { get; set; }
+
+    /// <summary>JSON-serialized <see cref="WebhookEventEnvelope"/> sent as the request body.</summary>
+    [BsonElement("payload")]
+    public required string Payload { get; set; }
+
+    [BsonElement("status")]
+    [BsonRepresentation(BsonType.String)]
+    public WebhookDeliveryStatus Status { get; set; } = WebhookDeliveryStatus.Pending;
+
+    [BsonElement("attempt_count")]
+    public int AttemptCount { get; set; }
+
+    [BsonElement("created_at")]
+    public DateTime CreatedAt { get; set; }
+
+    /// <summary>Earliest time this delivery may be (re)attempted.</summary>
+    [BsonElement("next_attempt_at")]
+    public DateTime NextAttemptAt { get; set; }
+
+    /// <summary>When set and in the future, this delivery is leased to <see cref="ClaimedBy"/>.</summary>
+    [BsonElement("lease_expires_at")]
+    public DateTime? LeaseExpiresAt { get; set; }
+
+    /// <summary>Identifier of the instance currently holding the lease.</summary>
+    [BsonElement("claimed_by")]
+    public string? ClaimedBy { get; set; }
+
+    [BsonElement("delivered_at")]
+    public DateTime? DeliveredAt { get; set; }
+
+    [BsonElement("last_error")]
+    public string? LastError { get; set; }
+}
--- a/XiansAi.Server.Src/Shared/Repositories/WebhookDeliveryRepository.cs
+++ b/XiansAi.Server.Src/Shared/Repositories/WebhookDeliveryRepository.cs
@@ -1,0 +1,156 @@
+using MongoDB.Driver;
+using Shared.Data;
+using Shared.Data.Models;
+using Shared.Utils;
+
+namespace Shared.Repositories;
+
+/// <summary>
+/// Persistence for the webhook outbox (collection <c>webhook_deliveries</c>). Provides an atomic
+/// claim so that, across multiple server instances, exactly one instance delivers each row.
+/// </summary>
+public interface IWebhookDeliveryRepository
+{
+    /// <summary>Inserts one or more pending delivery rows.</summary>
+    Task InsertManyAsync(IReadOnlyList<WebhookDelivery> deliveries);
+
+    /// <summary>
+    /// Atomically claims up to <paramref name="batchSize"/> deliveries that are due for
+    /// (re)attempt and not currently leased, transitioning them to <see cref="WebhookDeliveryStatus.Delivering"/>.
+    /// Also reclaims rows whose previous lease has expired (e.g. a crashed instance).
+    /// </summary>
+    Task<List<WebhookDelivery>> ClaimDueBatchAsync(string claimedBy, DateTime now, TimeSpan lease, int batchSize);
+
+    /// <summary>Marks a delivery as successfully delivered.</summary>
+    Task MarkDeliveredAsync(string id, DateTime deliveredAt);
+
+    /// <summary>Schedules a delivery for another attempt after <paramref name="nextAttemptAt"/>.</summary>
+    Task MarkRetryAsync(string id, int attemptCount, DateTime nextAttemptAt, string? error);
+
+    /// <summary>Marks a delivery as permanently failed.</summary>
+    Task MarkFailedAsync(string id, int attemptCount, string? error);
+}
+
+public class WebhookDeliveryRepository : IWebhookDeliveryRepository
+{
+    private const string CollectionName = "webhook_deliveries";
+
+    private readonly IMongoCollection<WebhookDelivery> _collection;
+    private readonly ILogger<WebhookDeliveryRepository> _logger;
+
+    public WebhookDeliveryRepository(IDatabaseService databaseService, ILogger<WebhookDeliveryRepository> logger)
+    {
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        var database = databaseService.GetDatabaseAsync().GetAwaiter().GetResult();
+        _collection = database.GetCollection<WebhookDelivery>(CollectionName);
+    }
+
+    public async Task InsertManyAsync(IReadOnlyList<WebhookDelivery> deliveries)
+    {
+        if (deliveries == null || deliveries.Count == 0)
+        {
+            return;
+        }
+
+        await MongoRetryHelper.ExecuteWithRetryAsync(async () =>
+        {
+            await _collection.InsertManyAsync(deliveries);
+        }, _logger, maxRetries: 3, baseDelayMs: 100, operationName: "InsertWebhookDeliveries");
+    }
+
+    public async Task<List<WebhookDelivery>> ClaimDueBatchAsync(string claimedBy, DateTime now, TimeSpan lease, int batchSize)
+    {
+        var claimed = new List<WebhookDelivery>();
+
+        // A row is claimable when it is due (next_attempt_at <= now) and not actively leased.
+        // Both Pending rows and stale Delivering rows (expired lease) are eligible, which lets a
+        // surviving instance recover deliveries abandoned by a crashed one.
+        var filter = Builders<WebhookDelivery>.Filter.And(
+            Builders<WebhookDelivery>.Filter.In(d => d.Status,
+                new[] { WebhookDeliveryStatus.Pending, WebhookDeliveryStatus.Delivering }),
+            Builders<WebhookDelivery>.Filter.Lte(d => d.NextAttemptAt, now),
+            Builders<WebhookDelivery>.Filter.Or(
+                Builders<WebhookDelivery>.Filter.Eq(d => d.LeaseExpiresAt, (DateTime?)null),
+                Builders<WebhookDelivery>.Filter.Lte(d => d.LeaseExpiresAt, now)));
+
+        var update = Builders<WebhookDelivery>.Update
+            .Set(d => d.Status, WebhookDeliveryStatus.Delivering)
+            .Set(d => d.ClaimedBy, claimedBy)
+            .Set(d => d.LeaseExpiresAt, now.Add(lease));
+
+        var options = new FindOneAndUpdateOptions<WebhookDelivery>
+        {
+            ReturnDocument = ReturnDocument.After,
+            Sort = Builders<WebhookDelivery>.Sort.Ascending(d => d.NextAttemptAt)
+        };
+
+        for (var i = 0; i < batchSize; i++)
+        {
+            var doc = await MongoRetryHelper.ExecuteWithRetryAsync(
+                async () => await _collection.FindOneAndUpdateAsync(filter, update, options),
+                _logger, maxRetries: 3, baseDelayMs: 100, operationName: "ClaimWebhookDelivery");
+
+            if (doc == null)
+            {
+                break;
+            }
+
+            claimed.Add(doc);
+        }
+
+        return claimed;
+    }
+
+    public async Task MarkDeliveredAsync(string id, DateTime deliveredAt)
+    {
+        var update = Builders<WebhookDelivery>.Update
+            .Set(d => d.Status, WebhookDeliveryStatus.Delivered)
+            .Set(d => d.DeliveredAt, deliveredAt)
+            .Set(d => d.LeaseExpiresAt, (DateTime?)null)
+            .Set(d => d.LastError, (string?)null);
+
+        await UpdateByIdAsync(id, update, "MarkWebhookDeliveryDelivered");
+    }
+
+    public async Task MarkRetryAsync(string id, int attemptCount, DateTime nextAttemptAt, string? error)
+    {
+        var update = Builders<WebhookDelivery>.Update
+            .Set(d => d.Status, WebhookDeliveryStatus.Pending)
+            .Set(d => d.AttemptCount, attemptCount)
+            .Set(d => d.NextAttemptAt, nextAttemptAt)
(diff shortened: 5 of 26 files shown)


hasith (commented):
## 🔍 Automated PR Review — PR #425: Enhance webhook functionality and configuration

**Verdict: REQUEST CHANGES** (advisory — non-blocking)

**Tier:** Escalated specialist review (webhook secrets, auth patterns, HMAC signing detected)
**Reviewers:** Code Quality · Security · Test Coverage · Performance

---

## Summary

This PR introduces a well-architected outbound webhook system using the transactional outbox pattern with atomic MongoDB lease claims, HMAC-SHA256 request signing, exponential backoff, and SSRF-mitigation headers. The core design is sound. However, there are **3 critical issues** that should be resolved before shipping to production, and **1 issue that breaks existing tests** which must be fixed before the branch can even compile cleanly.

---

## 🚨 Critical Issues

### C1 — Fire-and-forget over a Scoped `IWebhookDeliveryRepository` may silently drop deliveries
**File:** `XiansAi.Server.Src/Shared/Services/WebhookEventPublisher.cs:90`

`PersistDeliveriesAsync` is launched as a detached fire-and-forget task (`_ = PersistDeliveriesAsync(...)`) that holds a reference to `_repository`. Both `IWebhookDeliveryRepository` and `IWebhookEventPublisher` are registered as `AddScoped`, meaning the DI scope (and the repository wrapper) is disposed at the end of the HTTP request. The background task may outlive the scope, producing an `ObjectDisposedException` or a silent no-op write under high throughput.

**Fix:** Register both `WebhookDeliveryRepository` and `WebhookEventPublisher` as `AddSingleton` (the underlying `IMongoCollection<T>` is thread-safe). This is the simplest fix and is consistent with the codebase's existing singleton repository pattern.
```csharp
// In SharedServices.RegisterWebhookServices:
services.AddSingleton<IWebhookDeliveryRepository, WebhookDeliveryRepository>();
services.AddSingleton<IWebhookEventPublisher, WebhookEventPublisher>();
```

---

### C2 — N+1 MongoDB round-trips in `ClaimDueBatchAsync` — O(n) serial `FindOneAndUpdate` calls
**File:** `XiansAi.Server.Src/Shared/Repositories/WebhookDeliveryRepository.cs:87`

The dispatcher loops `batchSize` times calling `FindOneAndUpdate` sequentially. With the default `BatchSize=20` and `PollIntervalSeconds=5`, this is 20 sequential round-trips every 5 seconds. Under any backlog, the polling loop cannot keep up.

**Fix:** Use a two-step bulk approach — `UpdateMany` to stamp a unique `batchId` onto up to N due rows atomically, then `Find` by `batchId` in a single read (2 round-trips regardless of batch size).

---

### C3 — SSRF: `TryValidateUrl` only checks scheme, not destination host
**File:** `XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs:228`

`TryValidateUrl` allows any URL with an `https://` or `http://` scheme, including `https://169.254.169.254/latest/meta-data` (AWS IMDS), `https://127.0.0.1/admin`, or `https://10.0.0.1/internal`. A misconfigured (or socially-engineered) webhook subscription can exfiltrate internal webhook payloads containing PII (emails, user IDs, tenant data) to metadata services or internal infrastructure.

**Fix:** Add an explicit blocklist for loopback, RFC-1918 private ranges, and link-local before the scheme check:
```csharp
if (uri.Host.Equals("localhost", StringComparison.OrdinalIgnoreCase)
    || uri.Host.StartsWith("127.")
    || uri.Host.StartsWith("10.")
    || uri.Host.StartsWith("192.168.")
    || uri.Host.StartsWith("169.254.")
    || Regex.IsMatch(uri.Host, @"^172\.(1[6-9]|2\d|3[01])\."))
{
    error = $"Listener URL host '{uri.Host}' is a private/loopback address and is not permitted.";
    return false;
}
```

---

### C4 — `SecretVaultService` constructor change breaks existing test compilation
**File:** `XiansAi.Server.Src/Shared/Services/SecretVaultService.cs:111`

The `SecretVaultService` constructor now requires `IWebhookEventPublisher` as a fourth parameter, but the existing `SecretVaultServiceTests.cs` passes only 3 arguments and does not set up `GetByIdAsync` (now called unconditionally in `DeleteAsync` before publishing). This will cause a **compilation failure** and runtime `MockException` on the delete tests.

**Fix:** Update `SecretVaultServiceTests` to inject `Mock<IWebhookEventPublisher>` and add `_repo.Setup(r => r.GetByIdAsync(entity.Id)).ReturnsAsync(entity)` to the delete test cases.

---

## ⚠️ Warnings

### W1 — Claim index missing `lease_expires_at` — collection scan on lease-expiry recovery
**File:** `XiansAi.Server.Src/mongodb-indexes.yaml:534`

The `webhook_delivery_claim` index covers only `(status, next_attempt_at)`. The claim filter also includes `lease_expires_at IS NULL OR lease_expires_at <= now` for recovering stale `Delivering` rows. MongoDB cannot push this predicate to the index, so expired-lease recovery will scan every `Delivering` document as the collection grows.

**Fix:**
```yaml
- name: webhook_delivery_claim
  keys:
    status: asc
    next_attempt_at: asc
    lease_expires_at: asc
  partialFilterExpression:
    status: { $in: ["Pending", "Delivering"] }
```

---

### W2 — Sync-over-async `GetAwaiter().GetResult()` in `WebhookDeliveryRepository` constructor
**File:** `XiansAi.Server.Src/Shared/Repositories/WebhookDeliveryRepository.cs:44`

Calling `databaseService.GetDatabaseAsync().GetAwaiter().GetResult()` in the constructor synchronously blocks a thread-pool thread during DI resolution and risks deadlock in sync-over-async contexts.

**Fix:** Resolve the collection lazily with a `Task`-based initializer or use a static async factory method, consistent with avoiding this anti-pattern throughout the rest of the codebase.

---

### W3 — Serial delivery loop: 20 × 10s timeout = worst-case 200s per poll cycle
**File:** `XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs:128`

`DeliverAsync` is called sequentially inside `foreach (var delivery in claimed)`. All HTTP calls are independent I/O — they should be dispatched concurrently. With a 10-second request timeout and the default batch size of 20, worst-case throughput is ~6 deliveries/minute per dispatcher instance.

**Fix:** `await Task.WhenAll(claimed.Select(d => DeliverAsync(repository, d, stoppingToken)));`

---

### W4 — Graceful shutdown marks all in-flight deliveries `NextAttemptAt = UtcNow`, causing a restart burst
**File:** `XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs:176`

On `CancellationRequested`, every abandoned in-flight delivery is marked for immediate retry (`DateTime.UtcNow`). On a rolling restart with multiple instances, this can cause a sudden burst of concurrent claims for the same deliveries.

**Fix:** Use the existing next-attempt schedule or a short back-off rather than `DateTime.UtcNow` to spread the retry load.

---

### W5 — Email PII included in outbound webhook payloads with no redaction option
**File:** `XiansAi.Server.Src/Shared/Services/TenantParticipantUserService.cs:187`

User email addresses appear in `user.created`, `user.updated`, `user.tenant.added`, `user.role.changed`, and related events. These are serialized into the `payload` field, stored in MongoDB for 15 days, and transmitted to external listener URLs. There is no opt-out or PII-scrubbing configuration.

**Recommendation:** Add a per-subscription `IncludePii` boolean (defaulting to `false`) and strip email/name fields in `BuildDeliveries` unless explicitly opted in.

---

### W6 — `UserUpdated` (and related global events) published without `tenantId` — potential cross-tenant fan-out
**File:** `XiansAi.Server.Src/Shared/Services/GlobalUserAdminService.cs:209`

Three `PublishAsync` calls in `GlobalUserAdminService` omit the `tenantId` parameter. The resulting `tenant_id = null` delivery rows will match any subscription that doesn't filter by tenant, potentially broadcasting global admin user changes to subscriptions scoped for a specific tenant.

**Fix:** Pass `tenantId: null` explicitly with a comment documenting the intentional cross-tenant scope, and confirm subscriptions handle `null` tenant correctly.

---

### W7 — Double event emission on tenant enable/disable: `TenantEnabled`/`TenantDisabled` + `TenantUpdated`
**File:** `XiansAi.Server.Src/Shared/Services/TenantService.cs:419`

`UpdateTenant` emits `TenantEnabled`/`TenantDisabled` when the `Enabled` flag changes, then always calls `PersistTenantUpdate` which unconditionally emits `TenantUpdated`. A single toggle fires two events.

**Fix:** Either exclude enabled-state changes from `TenantUpdated`, or remove the separate `TenantEnabled`/`TenantDisabled` events and use a diff in the `TenantUpdated` payload.

---

### W8 — `ex.Message` from `HttpRequestException` / `SocketException` persisted to MongoDB
**File:** `XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs:180`

Raw exception messages often include destination hostnames, IP addresses, and OS error codes (e.g., `"Connection refused (127.0.0.1:80)"`). These land in `WebhookDelivery.LastError` in MongoDB and could surface through admin APIs or monitoring tools, leaking internal infrastructure topology.

**Fix:** Sanitize to a category-level message: `"HTTP transport error"` / `"Connection refused"` / `"Request timeout"` before persisting.

---

## 💡 Suggestions

### S1 — `AllowInsecureHttp=true` permits PII over plaintext without requiring a signing Secret
**File:** `XiansAi.Server.Src/Shared/Configuration/Options/WebhooksOptions.cs:47`

When `AllowInsecureHttp` is enabled, subscriptions can receive signed or unsigned PII payloads over plaintext HTTP. Consider requiring `Secret` to be set on any HTTP subscription to ensure at least the HMAC signature provides integrity even without TLS confidentiality.

---

### S2 — Redundant pre-check read before `UpsertAgentAsync` in `DefinitionsService`
**File:** `XiansAi.Server.Src/Features/AgentApi/Services/DefinitionsService.cs:315`

`GetByNameInternalAsync` is called immediately before `UpsertAgentAsync`, which itself already knows from its internal `InsertOne`/duplicate-key logic whether the row was created or updated. This adds an extra read round-trip to MongoDB on every agent registration/update.

**Fix:** Return a `(Agent agent, bool wasCreated)` tuple from `UpsertAgentAsync` and remove the pre-check read.

---

### S3 — Use `DateTimeOffset` instead of `DateTime` for `WebhookEventEnvelope.OccurredAt`
**File:** `XiansAi.Server.Src/Shared/Data/Models/WebhookDelivery.cs:243`

`DateTime` without timezone info is ambiguous for external listeners. `DateTimeOffset` is the preferred type for JSON interchange timestamps.

---

### S4 — No unit tests for `WebhookDispatcherService`, `WebhookEventPublisher`, or `MatchesEvent`
The webhook infrastructure adds ~726 lines of new security-sensitive production code (HMAC signing, SSRF prevention, atomic claim, retry scheduling) with zero corresponding unit tests. At minimum, add tests for: HMAC signature correctness, `TryValidateUrl` (good/bad URLs), retry boundary conditions, `MatchesEvent` wildcard/case-insensitive matching, and the disabled-early-exit path.

---

## ✅ Strengths

The overall architecture of this PR is well-conceived:
- **Outbox pattern** with atomic `FindOneAndUpdate` lease prevents duplicate delivery across instances
- **SSRF mitigations** on the `HttpClient`: `AllowAutoRedirect = false`, no cookies, no auto-decompression
- **HMAC-SHA256 request signing** with per-retry fresh timestamp prevents static replay
- **`WebhookEventTypes` constant catalog** is clean and discoverable
- **Publisher snapshot design** correctly captures ephemeral `ITenantContext` (actor, tenant) on the caller's thread before the scope ends
- **Startup validation** (`ValidateSubscriptionsAtStartup`) warns operators about missing secrets/invalid URLs
- **`OUTBOUND_WEBHOOKS.md`** is thorough and includes an HMAC verification recipe

---

_Review generated by Claude Code PR Reviewer · 4 specialists · PR #425_

hasith (commented):
Review posted — see full report in comments above. Verdict: REQUEST CHANGES (advisory).

XiansAi.Server.Src/Shared/Services/WebhookEventPublisher.cs
```diff
@@ -0,0 +1,185 @@
...
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Failed to build webhook deliveries for event {EventType}", eventType);
+            return Task.CompletedTask;
+        }
+
+        // Persist the outbox rows in the background. We deliberately do not await this: the write
+        // (with its retries) must not add latency to, or fail, the originating business operation.
+        // This trades a small durability window for isolation — publishing is best-effort.
+        _ = PersistDeliveriesAsync(deliveries, eventType);
```
hasith, line 90: 🚨 **CRITICAL — C1: Fire-and-forget over a Scoped `IWebhookDeliveryRepository` may silently drop deliveries**

`PersistDeliveriesAsync` is launched as a detached task while holding a reference to `_repository`, which is registered as `AddScoped`. The DI scope (and the repository wrapper object) is disposed at the end of the HTTP request. The background task may outlive the scope, producing an `ObjectDisposedException` or silent data loss under load.

**Fix:** Register both `WebhookDeliveryRepository` and `WebhookEventPublisher` as `AddSingleton` — the underlying `IMongoCollection<T>` is thread-safe and long-lived by design:
```csharp
// In SharedServices.RegisterWebhookServices:
services.AddSingleton<IWebhookDeliveryRepository, WebhookDeliveryRepository>();
services.AddSingleton<IWebhookEventPublisher, WebhookEventPublisher>();
```

XiansAi.Server.Src/Shared/Repositories/WebhookDeliveryRepository.cs
```diff
@@ -0,0 +1,156 @@
...
+            .Set(d => d.ClaimedBy, claimedBy)
+            .Set(d => d.LeaseExpiresAt, now.Add(lease));
+
+        var options = new FindOneAndUpdateOptions<WebhookDelivery>
+        {
+            ReturnDocument = ReturnDocument.After,
+            Sort = Builders<WebhookDelivery>.Sort.Ascending(d => d.NextAttemptAt)
+        };
+
+        for (var i = 0; i < batchSize; i++)
```
hasith, line 87: 🚨 **CRITICAL — C2: N+1 MongoDB round-trips — up to `batchSize` serial `FindOneAndUpdate` calls per poll cycle**

This loop issues 20 sequential round-trips to MongoDB every 5 seconds (default config). Under any backlog, the dispatcher cannot keep pace.

**Fix:** Use a two-step bulk approach — `UpdateMany` to stamp a unique `batchId` on up to N due rows atomically, then retrieve them with a single `Find` by `batchId`. This reduces 20 round-trips to 2:
```csharp
var batchId = Guid.NewGuid().ToString("N");
await _collection.UpdateManyAsync(filter,
    Builders<WebhookDelivery>.Update
        .Set(d => d.Status, WebhookDeliveryStatus.Delivering)
        .Set(d => d.ClaimedBy, claimedBy)
        .Set(d => d.ClaimBatchId, batchId)
        .Set(d => d.LeaseExpiresAt, now.Add(lease)));
var claimed = await _collection
    .Find(Builders<WebhookDelivery>.Filter.Eq(d => d.ClaimBatchId, batchId))
    .ToListAsync();
```

XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs
```diff
@@ -0,0 +1,280 @@
...
+            return (true, null);
+        }
+
+        return (false, $"HTTP {(int)response.StatusCode} {response.ReasonPhrase}");
+    }
+
+    /// <summary>
+    /// Validates that a listener URL is an absolute HTTPS URL (or HTTP when explicitly allowed).
+    /// </summary>
+    private bool TryValidateUrl(string url, out string? error)
```
hasith, line 228: 🚨 **CRITICAL — C3: SSRF — `TryValidateUrl` only checks scheme, not destination host**

This function allows `https://169.254.169.254/latest/meta-data` (AWS IMDS), `https://127.0.0.1/admin`, `https://10.0.0.1/internal`, etc. A misconfigured or socially-engineered webhook subscription can exfiltrate PII-containing payloads to internal infrastructure.

**Fix:** Add a host blocklist immediately after the URI parse:
```csharp
if (uri.Host.Equals("localhost", StringComparison.OrdinalIgnoreCase)
    || uri.Host.StartsWith("127.")
    || uri.Host.StartsWith("10.")
    || uri.Host.StartsWith("192.168.")
    || uri.Host.StartsWith("169.254.")
    || Regex.IsMatch(uri.Host, @"^172\.(1[6-9]|2\d|3[01])\."))
{
    error = $"Listener URL host '{uri.Host}' is a private/loopback address and is not permitted.";
    return false;
}
```

XiansAi.Server.Src/Shared/Services/SecretVaultService.cs
```diff
@@ -105,15 +105,18 @@ public class SecretVaultService : ISecretVaultService
 
     private readonly ISecretVaultRepository _repository;
     private readonly ISecretStoreProvider _secretStore;
+    private readonly IWebhookEventPublisher _webhookEventPublisher;
     private readonly ILogger<SecretVaultService> _logger;
 
     public SecretVaultService(
```
hasith, line 111: 🚨 **CRITICAL — C4: Constructor change breaks existing `SecretVaultServiceTests` compilation**

Adding `IWebhookEventPublisher` as a 4th constructor parameter breaks the existing test class, which still passes only 3 arguments. Additionally, `DeleteAsync` now calls `GetByIdAsync` unconditionally before publishing, but the tests use `MockBehavior.Strict` without setting up that method — causing a runtime `MockException`.

**Fix required in `SecretVaultServiceTests.cs`:**
```csharp
// Add to test setup:
var mockPublisher = new Mock<IWebhookEventPublisher>();
mockPublisher.Setup(p => p.PublishAsync(It.IsAny<string>(), It.IsAny<object>(), It.IsAny<string>()))
    .Returns(Task.CompletedTask);

// Add to delete test cases:
_repo.Setup(r => r.GetByIdAsync(entity.Id)).ReturnsAsync(entity);
```

XiansAi.Server.Src/mongodb-indexes.yaml
```diff
@@ -528,3 +528,23 @@ app_integrations:
       activation_name: asc
       name: asc
     unique: true
+
+webhook_deliveries:
+  # Drives the dispatcher's atomic claim query (status + due time).
+  - name: webhook_delivery_claim
```
hasith, line 534: ⚠️ **WARNING — W1: Claim index missing `lease_expires_at` — collection scan on lease-expiry recovery**

The claim query includes `lease_expires_at IS NULL OR lease_expires_at <= now` to reclaim stale `Delivering` rows, but the index only covers `(status, next_attempt_at)`. MongoDB cannot push the `lease_expires_at` predicate to the index, causing a full scan of all `Delivering` rows during recovery as the collection grows.

**Fix:**
```yaml
- name: webhook_delivery_claim
  keys:
    status: asc
    next_attempt_at: asc
    lease_expires_at: asc
  partialFilterExpression:
    status: { $in: ["Pending", "Delivering"] }
```

XiansAi.Server.Src/Shared/Repositories/WebhookDeliveryRepository.cs
```diff
@@ -0,0 +1,156 @@
...
+{
+    private const string CollectionName = "webhook_deliveries";
+
+    private readonly IMongoCollection<WebhookDelivery> _collection;
+    private readonly ILogger<WebhookDeliveryRepository> _logger;
+
+    public WebhookDeliveryRepository(IDatabaseService databaseService, ILogger<WebhookDeliveryRepository> logger)
+    {
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        var database = databaseService.GetDatabaseAsync().GetAwaiter().GetResult();
```
hasith, line 44: ⚠️ **WARNING — W2: Sync-over-async `GetAwaiter().GetResult()` in constructor**

Blocking a thread-pool thread during DI resolution is an anti-pattern that risks deadlock in sync-over-async contexts and wastes thread resources during startup.

**Fix:** Resolve the collection lazily — store `IDatabaseService` and call `await GetDatabaseAsync()` at the start of each operation, or use a `Lazy<Task<IMongoCollection<WebhookDelivery>>>` initialized once on first use.

XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs
```diff
@@ -0,0 +1,280 @@
...
+        var lease = TimeSpan.FromSeconds(Math.Max(5, _options.LeaseSeconds));
+        var batchSize = Math.Max(1, _options.BatchSize);
+
+        var claimed = await repository.ClaimDueBatchAsync(_instanceId, now, lease, batchSize);
+        if (claimed.Count == 0)
+        {
+            return;
+        }
+
+        foreach (var delivery in claimed)
```
hasith, line 128: ⚠️ **WARNING — W3: Serial delivery loop — worst-case 200s for a batch of 20**

Each `DeliverAsync` call is purely I/O-bound (HTTP POST). With the default `RequestTimeoutSeconds=10` and `BatchSize=20`, sequential processing gives a worst-case throughput of ~6 deliveries/minute per dispatcher instance. The calls are fully independent and should run concurrently.

**Fix:**
```csharp
await Task.WhenAll(
    claimed
        .TakeWhile(_ => !stoppingToken.IsCancellationRequested)
        .Select(d => DeliverAsync(repository, d, stoppingToken))
);
```

XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs
```diff
@@ -0,0 +1,280 @@
...
+                    delivery.Id, delivery.EventType, subscription.Name);
+                return;
+            }
+
+            await ScheduleRetryOrFailAsync(repository, delivery, attemptCount, detail);
+        }
+        catch (OperationCanceledException) when (stoppingToken.IsCancellationRequested)
+        {
+            // Shutting down: release the lease so another instance/attempt can pick it up promptly.
+            await repository.MarkRetryAsync(delivery.Id, delivery.AttemptCount, DateTime.UtcNow, "Dispatcher shutting down");
```
hasith, line 176: ⚠️ **WARNING — W4: Graceful shutdown marks all in-flight deliveries as immediately due, causing restart burst**

Setting `NextAttemptAt = DateTime.UtcNow` on shutdown means all abandoned deliveries become due simultaneously on the next startup. On a rolling restart with multiple instances this creates a thundering-herd claim attempt.

**Fix:** Use the delivery's existing `NextAttemptAt` (preserving the current schedule) or apply a short jittered delay rather than `UtcNow`, so retries are spread out after restart.

XiansAi.Server.Src/Shared/Services/TenantParticipantUserService.cs
```diff
@@ -180,6 +183,19 @@ public async Task<ServiceResult<TenantParticipantUser>> CreateAsync(string tenan
 
         var created = result.Data;
         var tenantRole = created.TenantRoles.FirstOrDefault(t => t.Tenant == tenantId);
+
+        await _webhookEventPublisher.PublishAsync(
```
hasith, line 187: ⚠️ **WARNING — W5: Email PII included in outbound webhook payloads with no redaction option**

User email addresses appear in `user.created`, `user.updated`, `user.tenant.added`, `user.role.changed`, and related events. These are serialized into MongoDB's `webhook_deliveries.payload` field (retained 15 days) and transmitted to external listener URLs. There is no per-subscription opt-out or PII-scrubbing configuration.

**Recommendation:** Add a per-subscription `IncludePii` boolean (default `false`) and strip `email`/`name` fields in `BuildDeliveries` unless explicitly opted in. This is particularly important for GDPR/data-residency compliance.

XiansAi.Server.Src/Shared/Services/GlobalUserAdminService.cs
```diff
@@ -203,6 +206,10 @@ public async Task<ServiceResult<GlobalUserDetail>> UpdateProfileAsync(string use
             await InvalidateCachesAsync(user);
             _logger.LogInformation("Global user {UserId} profile updated", LogSanitizer.Sanitize(userId));
 
+            await _webhookEventPublisher.PublishAsync(
```
hasith, line 209: ⚠️ **WARNING — W6: `UserUpdated` published without `tenantId` — potential cross-tenant fan-out**

The missing third argument means `tenant_id = null` in the delivery rows. Any subscription that doesn't filter by tenant will receive this event, potentially broadcasting global admin user changes to subscriptions scoped for a specific tenant.

**Fix:** Pass `tenantId: null` explicitly with a comment documenting the intentional global scope, and verify all subscription configurations handle `null` tenant correctly:
```csharp
await _webhookEventPublisher.PublishAsync(
    WebhookEventTypes.UserUpdated,
    new { userId = user.UserId, email = user.Email, name = user.Name },
    tenantId: null);  // global admin event — intentionally cross-tenant
```

XiansAi.Server.Src/Shared/Services/TenantService.cs
```diff
@@ -396,7 +412,17 @@ public async Task<ServiceResult<Tenant>> UpdateTenant(string id, UpdateTenantReq
             if (request.Timezone != null)
                 existingTenant.Timezone = request.Timezone;
 
-            return await PersistTenantUpdate(existingTenant, id);
+            var result = await PersistTenantUpdate(existingTenant, id);
+
+            if (result.IsSuccess && request.Enabled.HasValue && request.Enabled.Value != wasEnabled)
+            {
+                await _webhookEventPublisher.PublishAsync(
```
hasith, line 419: ⚠️ **WARNING — W7: Double event emission on tenant enable/disable toggle**

`UpdateTenant` emits `TenantEnabled`/`TenantDisabled` here when the `Enabled` flag changes, then always calls `PersistTenantUpdate` which unconditionally emits `TenantUpdated`. A single enable/disable toggle fires **two events** for the same state change.

**Fix:** Either (a) exclude `enabled`-only changes from triggering `TenantUpdated` in `PersistTenantUpdate`, or (b) remove the separate `TenantEnabled`/`TenantDisabled` events and encode the change in the `TenantUpdated` payload.

XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs
```diff
@@ -0,0 +1,280 @@
...
+        }
+        catch (Exception ex)
+        {
+            await ScheduleRetryOrFailAsync(repository, delivery, attemptCount, ex.Message);
```
hasith, line 180: ⚠️ **WARNING — W8: Raw `ex.Message` from network exceptions stored in MongoDB**

`HttpRequestException` and `SocketException` messages often include destination hostnames, IP addresses, and OS error codes (e.g., `"Connection refused (127.0.0.1:80)"`). These land in `WebhookDelivery.LastError` in MongoDB and may surface through admin APIs or monitoring dashboards, leaking internal infrastructure topology.

**Fix:** Sanitize to a category-level message before persisting:
```csharp
catch (HttpRequestException ex)
{
    var safeError = $"HTTP transport error ({ex.HttpRequestError}): connection failed";
    await ScheduleRetryOrFailAsync(repository, delivery, attemptCount, safeError);
}
catch (Exception ex)
{
    _logger.LogError(ex, "Unexpected error delivering webhook {DeliveryId}", delivery.Id);
    await ScheduleRetryOrFailAsync(repository, delivery, attemptCount, "Unexpected delivery error");
}
```

XiansAi.Server.Src/Shared/Configuration/Options/WebhooksOptions.cs
```diff
@@ -0,0 +1,105 @@
...
+    /// Maximum number of deliveries claimed and processed per poll iteration.
+    /// </summary>
+    public int BatchSize { get; set; } = 20;
+
+    /// <summary>
+    /// When false (default), listener URLs must use HTTPS. Set to true only for trusted
+    /// internal listeners reachable over plain HTTP. Delivering event payloads (which may
+    /// contain PII) over HTTP exposes them to network eavesdropping.
+    /// </summary>
+    public bool AllowInsecureHttp { get; set; } = false;
```
hasith, line 47: 💡 **SUGGESTION — S1: `AllowInsecureHttp=true` permits PII over plaintext without requiring a signing Secret**

When this flag is enabled, subscriptions can receive user-lifecycle payloads (emails, user IDs, tenant data) over plaintext HTTP with no HMAC signature. Consider requiring subscriptions using HTTP to have a `Secret` set, so the HMAC provides at least integrity protection even without TLS confidentiality:
```csharp
// In ValidateSubscriptionsAtStartup:
if (Uri.TryCreate(sub.Url, UriKind.Absolute, out var u) 
    && u.Scheme == Uri.UriSchemeHttp 
    && string.IsNullOrEmpty(sub.Secret))
{
    _logger.LogError("Subscription '{Name}' uses plain HTTP without a Secret — deliveries will be abandoned.", sub.Name);
}
```

XiansAi.Server.Src/Features/AgentApi/Services/DefinitionsService.cs
```diff
@@ -295,6 +310,11 @@ public async Task<IResult> CreateAgentAsync(CreateAgentRequest request)
                 statusCode: StatusCodes.Status409Conflict);
         }
 
+        // Determine whether this is a brand-new agent (so we only publish agent.registered once,
+        // rather than on every worker restart which re-upserts the same agent).
+        var existingAgent = await _agentRepository.GetByNameInternalAsync(
```
hasith, line 315: 💡 **SUGGESTION — S2: Redundant pre-check read before `UpsertAgentAsync` adds an extra MongoDB round-trip**

`GetByNameInternalAsync` is called here to detect whether the agent is new (for publishing `AgentRegistered`), but `UpsertAgentAsync` itself already knows whether it inserted or updated — it performs an `InsertOne` and catches a duplicate-key exception internally. This adds one extra read on every agent registration/update path.

**Fix:** Return a `(Agent agent, bool wasCreated)` tuple from `UpsertAgentAsync` and use that to decide which event to publish, eliminating the pre-check read entirely.

XiansAi.Server.Src/Shared/Data/Models/WebhookDelivery.cs
```diff
@@ -0,0 +1,318 @@
...
+
+    /// <summary>Unique id of the logical event; identical across all subscription deliveries.</summary>
+    [JsonPropertyName("eventId")]
+    public required string EventId { get; set; }
+
+    [JsonPropertyName("tenantId")]
+    public string? TenantId { get; set; }
+
+    [JsonPropertyName("occurredAt")]
+    public DateTime OccurredAt { get; set; }
```
hasith, line 243: 💡 **SUGGESTION — S3: Use `DateTimeOffset` instead of `DateTime` for `OccurredAt`**

`DateTime` without explicit timezone info is ambiguous for external webhook consumers. Listeners running in non-UTC timezones may misinterpret the value. `DateTimeOffset` is the recommended type for JSON interchange timestamps.

```csharp
[JsonPropertyName("occurredAt")]
public DateTimeOffset OccurredAt { get; set; }
```

hasith (commented):
## 🔍 Automated PR Review — PR #425: Enhance webhook functionality and configuration

**Verdict:** ⚠️ REQUEST CHANGES (advisory — non-blocking)

**Scope:** 26 files changed, 1,808 insertions, ~2,695 diff lines. New outbound webhook infrastructure: `WebhookEventPublisher`, `WebhookDispatcherService`, `WebhookDelivery` outbox model, `WebhookDeliveryRepository`, and event integrations across 15+ services.

**Review tier:** Specialist (security + test + performance + code quality) — triggered by webhook, HMAC, and API key surface areas.

---

## 🚨 Critical Issues (5)

### 1. Scoped repository used in fire-and-forget task — will silently drop events under load
**File:** `XiansAi.Server.Src/Shared/Services/WebhookEventPublisher.cs:90`

`_ = PersistDeliveriesAsync(deliveries, eventType)` detaches a background task that holds `IWebhookDeliveryRepository`, which is registered as `AddScoped` in `SharedConfiguration.cs`. When the HTTP request scope is disposed (which happens before the async I/O completes under any real load), the repository instance is disposed and the background task throws `ObjectDisposedException`, silently dropping the webhook delivery row.

**Fix:** Use `IServiceScopeFactory` to create a dedicated scope inside `PersistDeliveriesAsync` (the same pattern used correctly in `WebhookDispatcherService`), or register `WebhookDeliveryRepository` as `AddSingleton` — the underlying `IMongoCollection<T>` is thread-safe.

---

### 2. `KnowledgeService.DeleteById` never publishes `KnowledgeDeleted` event
**File:** `XiansAi.Server.Src/Shared/Services/KnowledgeService.cs:255`

All four other knowledge delete paths publish `WebhookEventTypes.KnowledgeDeleted`, but the primary `IResult DeleteById(string id)` API (used directly by admin/API controllers) silently skips the event. This is a logic inconsistency, not a style issue — a deployed listener relying on delete events will miss every deletion made via this path.

**Fix:** Add `await _webhookEventPublisher.PublishAsync(WebhookEventTypes.KnowledgeDeleted, new { id, ... }, tenantId)` after the successful `DeleteAsync` call on line 255.

---

### 3. `ClaimDueBatchAsync` issues N sequential MongoDB round-trips
**File:** `XiansAi.Server.Src/Shared/Repositories/WebhookDeliveryRepository.cs:87`

The loop issues `batchSize` sequential `FindOneAndUpdate` calls — 20 serialized network calls per poll cycle at the default batch size. Under load with fast delivery (events firing at high frequency), this creates 20+ blocking round-trips every 5 seconds, blocking the poll loop and degrading MongoDB for other operations. The lease window may expire before all documents are claimed.

**Fix:** Replace the loop with `UpdateMany` to stamp a unique batch-claim token on all matching documents, then `Find` by that token — 2 round-trips instead of N.

---

### 4. SSRF: `TryValidateUrl` does not block private/loopback IP ranges
**File:** `XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs:228`

The URL validator only checks the scheme (`https` vs `http`). A subscription URL of `https://169.254.169.254/latest/meta-data/` (AWS IMDS), `https://10.0.0.1/admin`, or `https://127.0.0.1:8080/internal` passes validation and the dispatcher will make outbound HTTP requests to those internal addresses. Even with operator-only configuration today, this is a significant SSRF risk if configuration injection or a future admin UI exposes subscriptions.

**Fix:** After `Uri.TryCreate`, resolve the hostname via `Dns.GetHostAddressesAsync` and reject any address in RFC-1918 private ranges (10.x, 172.16–31.x, 192.168.x), loopback (127.x, ::1), and link-local (169.254.x, fe80::). Re-validate at delivery time to guard against DNS rebinding.

---

### 5. `WebhookDispatcherService` dispatches claimed deliveries sequentially
**File:** `XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs:128`

The `foreach` loop in `ProcessDueDeliveriesAsync` dispatches one delivery at a time. With `BatchSize=20` and `RequestTimeoutSeconds=10` (defaults), one slow/timing-out listener can consume the full 200-second processing window — far beyond the 60-second `LeaseSeconds` default. This causes other dispatcher instances to reclaim and re-deliver the same messages (duplicate delivery).

**Fix:** Use `Task.WhenAll` with a `SemaphoreSlim` for bounded concurrency, and ensure `LeaseSeconds >= BatchSize × RequestTimeoutSeconds`.

---

## ⚠️ Warnings (10)

### 6. PII (email address) stored in webhook outbox and transmitted to external listeners
**File:** `XiansAi.Server.Src/Shared/Services/TenantParticipantUserService.cs` (user.created / user.updated events)

User-related event payloads include raw email addresses, stored plaintext in MongoDB `webhook_deliveries` (15-day TTL) and POSTed to external listener URLs. Those endpoints may not have equivalent data-protection controls to the primary user store.

**Fix:** Either gate email inclusion behind an explicit `includeEmail: true` opt-in in `WebhookSubscriptionOptions`, or replace raw email with a hashed/opaque user identifier in the transmitted payload.

---

### 7. `WebhookEventPublisher.PublishAsync` and `WebhookDispatcherService.DeliverAsync` have zero tests
**File:** `XiansAi.Server.Src/Shared/Services/WebhookEventPublisher.cs:57`
**File:** `XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs:139`

~800 lines of new production code (publisher, dispatcher, delivery model, repository) have no dedicated tests. The HMAC signing formula, exponential backoff boundaries, atomic-claim semantics, subscription fan-out, the disabled-gate short-circuit, and every error path are completely uncovered. A silent regression here breaks all listeners with no automated detection.

**Fix:** See Suggestions section for minimal test cases. At minimum add: HMAC golden-path test (pin the hex output), backoff boundary theory test, `MatchesEvent` logic test, and `PublishAsync` disabled/no-match gate tests.

---

### 8. `MaxAttempts` and `BatchSize` have no upper-bound clamping
**File:** `XiansAi.Server.Src/Shared/Configuration/Options/WebhooksOptions.cs:29`

An environment variable misconfiguration of `MaxAttempts=1000000` or `BatchSize=100000` would cause unbounded retry storms or O(N) sequential MongoDB calls per poll cycle, creating a self-inflicted DoS against the application.

**Fix:** Add `Math.Clamp(_options.MaxAttempts, 1, 25)` and `Math.Clamp(_options.BatchSize, 1, 200)` at consumption sites, or use `[Range]` data annotations with startup validation.

---

### 9. MongoDB claim index missing `lease_expires_at` field
**File:** `XiansAi.Server.Src/mongodb-indexes.yaml` (webhook_delivery_claim index)

The `ClaimDueBatchAsync` filter uses a `$or` on `lease_expires_at` (null or <= now), but the index only covers `{status, next_attempt_at}`. MongoDB must do an in-memory filter pass for the lease expiry condition on every matched row, which will degrade as `Delivering` rows accumulate under concurrent workers.

**Fix:** Extend the index to `{status: asc, next_attempt_at: asc, lease_expires_at: asc}`.

---

### 10. `TenantService` emits duplicate events on enable/disable
**File:** `XiansAi.Server.Src/Shared/Services/TenantService.cs:546`

When a tenant's `Enabled` field is toggled, both a `tenant.updated` event (from `PersistTenantUpdate`) and a `tenant.enabled`/`tenant.disabled` event are published for the same user action. Listeners receive two events and may process the state change twice.

**Fix:** Either suppress `tenant.updated` inside `PersistTenantUpdate` when the change is exclusively an enable/disable toggle, or consolidate into a single conditional event.

---

### 11. `ComputeSignature` allocates a new `HMACSHA256` and encodes payload bytes on every delivery
**File:** `XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs:274`

A new `HMACSHA256` instance (key schedule setup) is constructed per delivery call. The `$"{timestamp}.{delivery.Payload}"` intermediate string doubles the payload memory footprint before encoding. Under high throughput this creates significant per-delivery allocation pressure.

**Fix:** Cache HMAC key bytes per subscription (secret is static). Use `HMACSHA256.HashData(keyBytes, payloadBytes)` (.NET 6+) to avoid instance allocation. Avoid the concatenated string by writing timestamp + payload bytes directly into a `Span<byte>`.

---

### 12. `IWebhookDeliveryRepository.InsertManyAsync` scope and missing startup clamping on `GetDatabaseAsync().GetAwaiter().GetResult()`
**File:** `XiansAi.Server.Src/Shared/Repositories/WebhookDeliveryRepository.cs:44`

Constructor calls `databaseService.GetDatabaseAsync().GetAwaiter().GetResult()` — a sync-over-async that can deadlock in ASP.NET Core synchronization contexts on cold-start, especially if the MongoDB connection is slow. Combined with `AddScoped` registration, this blocking call runs per HTTP request scope.

**Fix:** Register as `AddSingleton` (thread-safe `IMongoCollection<T>` is long-lived), or use a `Lazy<Task<IMongoCollection<T>>>` initializer.

---

### 13. `GlobalUserAdminService` publishes user events with `tenantId: null` inconsistently
**File:** `XiansAi.Server.Src/Shared/Services/GlobalUserAdminService.cs`

All `PublishAsync` calls in this service omit `tenantId`, meaning `webhook_deliveries` rows have `null` in the tenant field. This is intentional for global operations, but breaks tenant-scoped audit queries and is inconsistent with all other call sites that pass an explicit tenant.

**Fix:** Document the intent with `tenantId: null` explicitly, or use a sentinel value (e.g. `"*"`) for global events and update queries accordingly.

---

### 14. `DefinitionsService` makes an extra `GetByNameInternalAsync` read before every upsert
**File:** `XiansAi.Server.Src/Features/AgentApi/Services/DefinitionsService.cs:115`

A pre-upsert read to detect new vs. existing agents fires on every agent worker startup / re-registration — a hot path. This doubles the MongoDB reads on agent connect.

**Fix:** Return an `IsNewlyCreated` flag from `UpsertAgentAsync` (inspect `result.IsUpsert` / `MatchedCount == 0`) and eliminate the pre-check read.

---

### 15. Weak HMAC secret placeholder in `.env.example` is indistinguishable from a real value
**File:** `XiansAi.Server.Src/.env.example:27`

`Webhooks__Subscriptions__0__Secret=shared-secret-for-hmac` looks like a real value and operators may copy it without substituting. CWE-521: Weak Password Requirements.

**Fix:** Replace with `=REPLACE_WITH_MIN_32_CHAR_RANDOM_SECRET` and add startup validation in `ValidateSubscriptionsAtStartup` rejecting secrets < 32 bytes.

---

## 💡 Suggestions (5)

| # | File | Line | Issue |
|---|---|---|---|
| S1 | `WebhooksOptions.cs` | 75 | Add `[JsonIgnore]` + `[DebuggerBrowsable(Never)]` to `Secret` property to prevent accidental diagnostic leakage |
| S2 | `WebhookDelivery.cs` | ~75 | Consider field-level encryption or PII scrubbing in stored `Payload` (15-day TTL with raw email/name) |
| S3 | `docs/OUTBOUND_WEBHOOKS.md` | HMAC section | Add language-specific `CryptographicOperations.FixedTimeEquals` example for constant-time comparison |
| S4 | `ApiKeyService.cs` | ~1164 | `ApiKeyRevoked` payload passes `name = existingKey?.Name` — may be `null` if key not found before deletion |
| S5 | `WebhookDeliveryRepository.cs` | ~87 | Secondary: even before the bulk-claim fix, the `ClaimDueBatchAsync` loop (finding S3) should have a configurable concurrency cap |

---

## ✅ Positive Observations

- **Outbox pattern implementation is architecturally sound** — atomic lease-based claiming via `FindOneAndUpdate`, exponential backoff, TTL cleanup, and crash-recovery are all correct in design.
- **SSRF partial mitigations are present** — `SocketsHttpHandler` disables redirects, cookies, and decompression, which are thoughtful defaults.
- **`ValidateSubscriptionsAtStartup` logging** — operators see misconfiguration at startup rather than at first delivery failure. Excellent DX.
- **`WebhookEventTypes` catalog** — comprehensive XML-doc'd static class makes event discovery trivial.
- **Consistent null-guard pattern** (`?? throw new ArgumentNullException(...)`) applied uniformly across all modified services.
- **`Truncate` helper** in `WebhookDeliveryRepository` prevents unbounded document growth from verbose exception messages.
- **`WebhookEventPublisher` actor snapshot** — correctly captures the authenticated context on the calling thread before the fire-and-forget, avoiding a race on a disposed HTTP context.

---

## Summary

| Dimension | Status |
|---|---|
| Code Quality | ⚠️ Missing webhook events in `KnowledgeService.DeleteById`; dual-event on tenant enable/disable |
| Security | 🚨 SSRF via unvalidated private IPs; PII in outbound payloads; scoped repo in fire-and-forget |
| Test Coverage | 🚨 ~0% coverage on all new core classes; critical HMAC/backoff paths completely unverified |
| Performance | 🚨 Sequential dispatch + N sequential MongoDB claims = lease expiry + duplicate delivery under load |

The architecture is solid and the outbox pattern is well-designed. The issues above are concentrated in three actionable areas: (1) fix the scoped DI lifetime and fire-and-forget pattern, (2) add private-IP blocking to the URL validator, and (3) add tests for the HMAC signing path, backoff logic, and `MatchesEvent` predicate.

---
*🤖 Review generated by Claude Code PR Review — [claude.ai/claude-code](https://claude.ai/claude-code)*

XiansAi.Server.Src/Shared/Services/WebhookEventPublisher.cs
hasith, line 90: **🚨 CRITICAL — Scoped repository used in fire-and-forget task**

`IWebhookDeliveryRepository` is registered as `AddScoped` in `SharedConfiguration.cs`. When the HTTP request scope ends (which happens before async I/O completes under any real load), the DI container disposes the repository instance. This fire-and-forget task then runs with a disposed repository and silently drops the webhook delivery row.

**Fix:** Use `IServiceScopeFactory` to create a dedicated scope inside `PersistDeliveriesAsync` (the same pattern used correctly in `WebhookDispatcherService`), or register `WebhookDeliveryRepository` as `AddSingleton` — the underlying `IMongoCollection<T>` is thread-safe.

```csharp
private async Task PersistDeliveriesAsync(IReadOnlyList<WebhookDelivery> deliveries, string eventType)
{
    using var scope = _scopeFactory.CreateScope();
    var repo = scope.ServiceProvider.GetRequiredService<IWebhookDeliveryRepository>();
    await repo.InsertManyAsync(deliveries);
}
```

XiansAi.Server.Src/Shared/Services/KnowledgeService.cs
```diff
@@ -317,6 +320,11 @@ public async Task<IResult> DeleteAllVersions(DeleteAllVersionsRequest request)
         if (!result)
             return Results.NotFound("Knowledge not found or could not be deleted");
 
+        await _webhookEventPublisher.PublishAsync(
+            WebhookEventTypes.KnowledgeDeleted,
```
hasith, line 324: **🚨 CRITICAL — `DeleteById` never publishes `KnowledgeDeleted` event (gap relative to this PR)**

This PR correctly adds `KnowledgeDeleted` publishing to `DeleteAllVersions` (here) and `DeleteByIdForTenantAsync`, but the primary `IResult DeleteById(string id)` method (line ~214) — used directly by API controllers — was not updated. Any listener relying on delete notifications will silently miss every deletion made via that path.

**Fix:** Add the publish call after the successful `DeleteAsync` call inside `DeleteById`:
```csharp
await _webhookEventPublisher.PublishAsync(
    WebhookEventTypes.KnowledgeDeleted,
    new { id, agent = knowledge.Agent, tenantId = knowledge.TenantId },
    knowledge.TenantId);
```

XiansAi.Server.Src/Shared/Repositories/WebhookDeliveryRepository.cs
hasith, line 87: **🚨 CRITICAL — N sequential `FindOneAndUpdate` round-trips (one per batch slot)**

This loop issues `batchSize` serialized MongoDB round-trips per poll cycle. At the default batch of 20, that is 20 blocking network calls before the dispatcher can start delivering. Under the default 5-second polling interval and 60-second lease, a slow MongoDB (or network blip) can cause the batch to stall and lease windows to expire — triggering redundant reclaims and duplicate deliveries by other instances.

**Fix:** Replace with 2 round-trips: one `UpdateMany` to stamp a unique `claimed_by` token on all matching documents, followed by one `Find` to retrieve them:
```csharp
var batchToken = Guid.NewGuid().ToString();
await _collection.UpdateManyAsync(filter, Builders<WebhookDelivery>.Update
    .Set(d => d.ClaimedBy, batchToken)
    .Set(d => d.LeaseExpiresAt, now.Add(lease)));
return await _collection.Find(d => d.ClaimedBy == batchToken).ToListAsync();
```

XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs
hasith, line 228: **🚨 CRITICAL — SSRF: `TryValidateUrl` does not block private/loopback IP ranges**

Only the URL scheme (`https` vs `http`) is validated. URLs like `https://169.254.169.254/latest/meta-data/` (cloud IMDS), `https://10.0.0.1/admin`, or `https://127.0.0.1:8080/internal` pass validation and the dispatcher will make outbound HTTP requests to internal infrastructure.

**Fix:** After `Uri.TryCreate`, resolve the hostname and reject private/reserved ranges:
```csharp
var addresses = await Dns.GetHostAddressesAsync(uri.Host);
foreach (var addr in addresses)
{
    if (IsPrivateOrReserved(addr)) // RFC-1918, loopback, link-local, APIPA
    {
        error = $"Listener URL resolves to a reserved address ({addr})";
        return false;
    }
}
```
Also re-validate at delivery time (not just startup) to guard against DNS rebinding attacks.

XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs
hasith, line 128: **🚨 CRITICAL — Sequential delivery loop: one slow listener blocks the entire batch**

With `BatchSize=20` and `RequestTimeoutSeconds=10` (defaults), a single timing-out listener endpoint consumes up to 200 seconds, far exceeding `LeaseSeconds=60`. Other dispatcher instances will reclaim the same deliveries and re-deliver them, causing duplicates.

**Fix:** Dispatch concurrently with bounded parallelism:
```csharp
var semaphore = new SemaphoreSlim(Math.Min(claimed.Count, 10));
await Task.WhenAll(claimed.Select(async delivery =>
{
    await semaphore.WaitAsync(stoppingToken);
    try { await DeliverAsync(repository, delivery, stoppingToken); }
    finally { semaphore.Release(); }
}));
```
Also ensure `LeaseSeconds >= BatchSize × RequestTimeoutSeconds` (currently 60 < 200).

XiansAi.Server.Src/Shared/Services/TenantParticipantUserService.cs
```diff
@@ -296,6 +333,37 @@ public async Task<ServiceResult<TenantParticipantUser>> UpdateAsync(
 
             await InvalidateCachesAsync(userId, tenantId);
 
+            if (profileChanged)
+            {
+                await _webhookEventPublisher.PublishAsync(
+                    WebhookEventTypes.UserUpdated,
+                    new { userId = user.UserId, email = user.Email, name = user.Name, tenantId },
```
hasith, line 340: **⚠️ WARNING — PII (email address) transmitted to external listeners and stored in outbox**

Raw email addresses are included in webhook payloads, persisted plaintext in `webhook_deliveries` for 15 days, and POSTed to external listener URLs. Those endpoints may not have equivalent data-protection controls to the primary user store (GDPR/data-processor implications).

**Fix:** Either gate email inclusion behind an explicit `includeEmail: true` opt-in in `WebhookSubscriptionOptions`, or replace raw email with an opaque `userId` in the transmitted payload. If email must be included, document that listener URLs are PII processors requiring appropriate data-protection agreements.

XiansAi.Server.Src/Shared/Services/WebhookEventPublisher.cs
```diff
@@ -0,0 +1,185 @@
...
+        ITenantContext tenantContext,
+        ILogger<WebhookEventPublisher> logger)
+    {
+        _options = options ?? throw new ArgumentNullException(nameof(options));
+        _repository = repository ?? throw new ArgumentNullException(nameof(repository));
+        _tenantContext = tenantContext ?? throw new ArgumentNullException(nameof(tenantContext));
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+    }
+
+    public Task PublishAsync(string eventType, object? data, string? tenantId = null)
```
hasith, line 57: **⚠️ WARNING — Zero test coverage on all new core webhook classes**

`WebhookEventPublisher`, `WebhookDispatcherService`, `WebhookDeliveryRepository`, and `WebhookSubscriptionOptions.MatchesEvent` have no dedicated tests (~800 lines of new production code, ~0% covered). The HMAC signing formula, exponential backoff boundaries, subscription fan-out, and the disabled-gate short-circuit are all completely unverified.

Minimum test set to add:
- **HMAC golden path:** pin the exact hex output for known secret + payload — prevents silent algorithm changes
- **Backoff boundary `[Theory]`:** assert seconds for attempts 1–6+
- **`MatchesEvent` `[Theory]`:** disabled / empty / wildcard / case-insensitive / no-match
- **`PublishAsync` disabled gate:** verify `InsertManyAsync` is never called when `Enabled=false`
- **Abandoned delivery:** verify `MarkFailedAsync` is called when subscription no longer exists

XiansAi.Server.Src/Shared/Configuration/Options/WebhooksOptions.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    /// <summary>
+    /// Per-request HTTP timeout when POSTing to a listener URL.
+    /// </summary>
+    public int RequestTimeoutSeconds { get; set; } = 10;
+
+    /// <summary>
+    /// Maximum number of delivery attempts before a delivery is marked as failed.
+    /// </summary>
+    public int MaxAttempts { get; set; } = 5;
```
hasith, line 29: **⚠️ WARNING — `MaxAttempts` and `BatchSize` have no upper-bound clamping**

An environment variable misconfiguration (`MaxAttempts=1000000` or `BatchSize=100000`) would cause unbounded retry storms or N sequential MongoDB `FindOneAndUpdate` calls per poll cycle, creating a self-inflicted DoS.

**Fix:** Add `[Range]` data annotations or clamp at consumption sites:
```csharp
var batchSize = Math.Clamp(_options.BatchSize, 1, 200);
var maxAttempts = Math.Clamp(_options.MaxAttempts, 1, 25);
```

XiansAi.Server.Src/Shared/Services/TenantService.cs
```diff
@@ -515,6 +541,19 @@ private async Task<ServiceResult<Tenant>> PersistTenantUpdate(Tenant existingTen
             _logger.LogWarning("Skipping tenant cache invalidation: Tenant {Id} has null or empty TenantId", LogSanitizer.Sanitize(id));
 
         _logger.LogInformation("Updated tenant with ID {Id}", LogSanitizer.Sanitize(id));
+
+        await _webhookEventPublisher.PublishAsync(
```
hasith, line 545: **⚠️ WARNING — Dual webhook events fired for a single enable/disable toggle**

When `request.Enabled` is set, `PersistTenantUpdate` publishes `tenant.updated` (line 545-548), and then the caller also publishes `tenant.enabled` or `tenant.disabled` (line 419-421). Listeners receive two events for a single user action and may process the state change twice.

**Fix:** Either suppress `tenant.updated` inside `PersistTenantUpdate` when the change is exclusively an enable/disable toggle, or consolidate to a single conditional event. At minimum add a comment if the dual-event behavior is intentional.

XiansAi.Server.Src/Shared/Services/WebhookDispatcherService.cs
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    /// <summary>Exponential backoff capped at 5 minutes: 10s, 20s, 40s, 80s, ...</summary>
+    private static TimeSpan ComputeBackoff(int attemptCount)
+    {
+        var seconds = 10d * Math.Pow(2, Math.Max(0, attemptCount - 1));
+        return TimeSpan.FromSeconds(Math.Min(300, seconds));
+    }
+
+    private static string ComputeSignature(string secret, string payload)
```
hasith, line 274: **⚠️ WARNING — New `HMACSHA256` instance + double byte-array allocation on every delivery**

`new HMACSHA256(Encoding.UTF8.GetBytes(secret))` reconstructs the key schedule per call. The `$"{timestamp}.{delivery.Payload}"` intermediate string doubles the payload in memory before encoding. Under high delivery volume this creates significant per-call allocation pressure.

**Fix:**
- Cache HMAC key bytes per subscription name (secret is static at runtime)
- Use `HMACSHA256.HashData(keyBytes, payloadBytes)` (.NET 6+) to avoid instance allocation
- Avoid the intermediate concatenated string using `Span<byte>` or by writing prefix + payload bytes directly into a buffer

XiansAi.Server.Src/Shared/Repositories/WebhookDeliveryRepository.cs
hasith, line 44: **⚠️ WARNING — Sync-over-async (`GetAwaiter().GetResult()`) in constructor + `AddScoped` registration**

`databaseService.GetDatabaseAsync().GetAwaiter().GetResult()` blocks the calling thread synchronously. Combined with `AddScoped` registration, this blocking call executes on every new HTTP request scope. On cold start or under a synchronization context (classic ASP.NET, test runners), this can deadlock.

**Fix:** Register as `AddSingleton` — the underlying `IMongoCollection<T>` is thread-safe and long-lived. A singleton eliminates both the per-request blocking call and the scoped-lifetime issue with the fire-and-forget task in `WebhookEventPublisher`.

XiansAi.Server.Src/Features/AgentApi/Services/DefinitionsService.cs
hasith, line 315: **⚠️ WARNING — Extra MongoDB read on every agent registration (hot path)**

`GetByNameInternalAsync` is called before every `UpsertAgentAsync` solely to detect whether the agent is new (for conditional webhook publishing). Agent workers re-register on every startup, so this extra read fires on every worker connect — doubling the MongoDB reads on a hot path.

**Fix:** Have `UpsertAgentAsync` return an `IsNewlyCreated` flag by inspecting the upsert result (`result.IsUpsert == true` / `MatchedCount == 0`), then publish `AgentRegistered` conditionally based on that flag — eliminating the pre-check read entirely.

XiansAi.Server.Src/.env.example
```diff
@@ -128,6 +128,31 @@ Logging__LogLevel__Microsoft.AspNetCore=Warning
...
+#Webhooks__RequestTimeoutSeconds=10
+#Webhooks__MaxAttempts=5
+#Webhooks__LeaseSeconds=60
+#Webhooks__BatchSize=20
+# Require HTTPS listener URLs (recommended). Set true only for trusted internal http listeners.
+#Webhooks__AllowInsecureHttp=false
+# One entry per listener. EventTypes empty or "*" means all events.
+#Webhooks__Subscriptions__0__Name=crm-listener
+#Webhooks__Subscriptions__0__Url=https://example.com/hooks/xians
+#Webhooks__Subscriptions__0__Secret=shared-secret-for-hmac
```
hasith, line 149: **⚠️ WARNING — Weak HMAC secret placeholder (CWE-521)**

`shared-secret-for-hmac` looks like a real value. Operators copying example configurations without substituting placeholders is a well-documented misconfiguration risk — any listener verifying HMAC signatures using this value provides no security.

**Fix:** Replace with an unambiguously unusable placeholder:
```
#Webhooks__Subscriptions__0__Secret=REPLACE_WITH_MIN_32_CHAR_RANDOM_SECRET
```
Also consider adding startup validation in `ValidateSubscriptionsAtStartup` to reject secrets shorter than 32 bytes.

XiansAi.Server.Src/mongodb-indexes.yaml
```diff
@@ -528,3 +528,23 @@ app_integrations:
...
+    keys:
+      status: asc
+      next_attempt_at: asc
```
hasith, line 537: **⚠️ WARNING — Claim index missing `lease_expires_at` field**

`ClaimDueBatchAsync` filters on `{ status, next_attempt_at, lease_expires_at }` (the $or condition for null or expired lease), but this index only covers `{status, next_attempt_at}`. MongoDB must apply an in-memory pass to filter the `lease_expires_at` condition on every matched row. As `Delivering` rows accumulate under concurrent workers, this in-memory filter will degrade performance.

**Fix:** Extend the index:
```yaml
keys:
  status: asc
  next_attempt_at: asc
  lease_expires_at: asc
```

XiansAi.Server.Src/Shared/Configuration/Options/WebhooksOptions.cs
```diff
@@ -0,0 +1,105 @@
...
+    /// <summary>
+    /// Destination URL that receives the HTTP POST for each matching event.
+    /// </summary>
+    public string Url { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Optional shared secret. When set, an HMAC-SHA256 signature of the request body is sent
+    /// in the <c>X-Xians-Signature</c> header so the listener can verify authenticity.
+    /// </summary>
+    public string? Secret { get; set; }
```
hasith, line 75: **💡 SUGGESTION — Add `[JsonIgnore]` and `[DebuggerBrowsable(Never)]` to `Secret` property**

The `Secret` property holds the HMAC signing key and is bound from `IConfiguration`. Without these attributes, it can appear in: health-check/configuration-dump endpoints, diagnostic tooling that serializes the options graph, and exception messages. 

**Fix:**
```csharp
[JsonIgnore]
[DebuggerBrowsable(DebuggerBrowsableState.Never)]
public string? Secret { get; set; }
```

hasith (commented):
**Review verdict: ⚠️ REQUEST CHANGES (advisory)**

5 critical issues and 10 warnings found — see the detailed review comment above and inline annotations for specific locations. The most urgent items before this ships to production:

1. 🚨 Fix the scoped-repository fire-and-forget lifetime bug () — will silently drop events under load
2. 🚨 Add private-IP-range blocking to  — SSRF risk
3. 🚨 Make delivery dispatch concurrent (or ensure ) — prevents duplicate delivery
4. 🚨 Add tests for HMAC signing, backoff logic, and  — currently 0% coverage on all new core classes
5. ⚠️ Add  publish to 

_🤖 Claude Code PR Review — [claude.ai/claude-code](https://claude.ai/claude-code)_

hasith (commented):
**Review verdict: REQUEST CHANGES (advisory)**

5 critical issues and 10 warnings found. See the detailed review comment above and inline annotations. Most urgent items before production:

1. CRITICAL — Fix scoped-repository fire-and-forget lifetime bug in WebhookEventPublisher — will silently drop events under load
2. CRITICAL — Add private-IP-range blocking to TryValidateUrl (SSRF risk)
3. CRITICAL — Make delivery dispatch concurrent, or set LeaseSeconds >= BatchSize times RequestTimeoutSeconds
4. CRITICAL — Add tests for HMAC signing, backoff logic, and MatchesEvent (0% coverage on all new core classes)
5. WARNING — Add KnowledgeDeleted publish to KnowledgeService.DeleteById

_Review by Claude Code PR Review_

PR #446, opened 2026-07-12
feat(api): implement case-insensitive tenant ID handling

hasith:
- Enhanced tenant creation and lookup processes to ensure tenant IDs are unique regardless of casing, preventing duplicates.
- Updated `BootstrapService` to use the resolved tenant ID from the result of the tenant existence check.
- Modified `TenantService` to reject tenant creation requests with duplicate IDs that differ only in case, improving data integrity.
- Added integration test to verify that creating a tenant with a duplicate ID in different casing returns a BadRequest response.

These changes improve the robustness of tenant management in the Admin API.

### DIFF
--- a/XiansAi.Server.Src/Shared/Services/TenantService.cs
+++ b/XiansAi.Server.Src/Shared/Services/TenantService.cs
@@ -637,6 +637,16 @@ public class TenantService : ITenantService
             var validatedTenant = tenant.SanitizeAndValidate();
             _logger.LogInformation("After sanitization, CreatedBy: {CreatedBy}", LogSanitizer.Sanitize(validatedTenant.CreatedBy));
 
+            // Tenant ids are unique ignoring case: reject "MyTenant" when "mytenant" already exists.
+            // The unique index on tenant_id remains the backstop for exact-case duplicates under races.
+            var duplicateTenant = await _tenantRepository.GetByTenantIdCaseInsensitiveAsync(validatedTenant.TenantId);
+            if (duplicateTenant != null)
+            {
+                _logger.LogWarning("Tenant ID {TenantId} already exists (case-insensitive match with {ExistingTenantId})",
+                    LogSanitizer.Sanitize(validatedTenant.TenantId), LogSanitizer.Sanitize(duplicateTenant.TenantId));
+                return ServiceResult<TenantCreatedResult>.BadRequest("A tenant with this ID already exists.");
+            }
+
             // Persist with Secret metadata values encrypted.
             validatedTenant.Metadata = _metadataProtector.Protect(validatedTenant.Metadata, validatedTenant.TenantId);
 
--- a/XiansAi.Server.Src/Features/AdminApi/Services/BootstrapService.cs
+++ b/XiansAi.Server.Src/Features/AdminApi/Services/BootstrapService.cs
@@ -94,11 +94,14 @@ public class BootstrapService : IBootstrapService
 
             // 4. Ensure an enabled tenant exists
             var tenantResult = await EnsureEnabledTenantAsync(resolvedTenantId, email);
-            if (!tenantResult.IsSuccess)
+            if (!tenantResult.IsSuccess || tenantResult.Data == null)
             {
                 return ServiceResult<BootstrapResponse>.Failure(
                     tenantResult.ErrorMessage ?? "Failed to ensure tenant", tenantResult.StatusCode);
             }
+
+            // Use the stored tenant id: an existing tenant may match the requested id with different casing.
+            resolvedTenantId = tenantResult.Data.TenantId;
 
             // 5. Create the SysAdmin user (CreateNewUser sets IsSysAdmin = true because no users exist)
             var createUserResult = await _userManagementService.CreateNewUser(new UserDto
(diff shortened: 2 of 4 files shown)


hasith (commented):
## PR Review Report

**PR:** feat(api): implement case-insensitive tenant ID handling
**Author:** Hasith Yaggahavita
**Files Changed:** 4 | **+69** / **-2**
**Verdict:** NEEDS DISCUSSION

---

### Summary
This PR adds case-insensitive tenant ID uniqueness enforcement at bootstrap and at `TenantService.CreateTenant`, backed by a new `GetByTenantIdCaseInsensitiveAsync` repository method that uses a MongoDB secondary-strength collation. The read-side implementation is clean and the new integration test correctly covers the happy-path duplicate-rejection scenario. However, the write-side guarantee has a check-then-act race: the existing unique index on `tenant_id` has no collation, so it is case-sensitive and does not back up the new case-insensitive check under concurrent requests.

---

No critical issues found.

---

### Warnings (Should Fix)

- [ ] `XiansAi.Server.Src/Shared/Services/TenantService.cs:642` — Check-then-act race: the case-insensitive duplicate check is not enforced atomically by the database, so concurrent `CreateTenant` calls with different casing (e.g. `MyTenant` / `mytenant`) can both pass the check and both insert, defeating the uniqueness guarantee this PR is meant to add.

---

- [ ] `XiansAi.Server.Src/Features/AdminApi/Services/BootstrapService.cs:182` — `EnsureEnabledTenantAsync` has the same check-then-act pattern as `TenantService.CreateTenant`. Lower risk since bootstrap is typically a single-admin, low-concurrency path, but worth applying the same fix for consistency if one is added.

---

### Review Details

#### Code Quality
Changes are small, well-commented, and follow existing patterns in the repository (retry-wrapped Mongo calls, `ServiceResult<T>` conventions, sanitized logging via `LogSanitizer`). The new repository method and its XML doc comment are clear.

#### Security
No injection or authz issues. Inputs are sanitized before reaching the new repository method, and log statements correctly use `LogSanitizer.Sanitize`. The one substantive issue is the concurrency race noted above, which is a data-integrity concern rather than a classic vulnerability.

#### Test Coverage
The new integration test (`CreateTenant_WithSameTenantIdInDifferentCase_ReturnsBadRequest`) exercises the intended sequential-request behavior well. There is no test for the concurrent-request race condition described above (understandably hard to test deterministically), and no test covering the bootstrap path's case-insensitive reuse behavior.

#### Performance
No notable performance concerns; the new lookup is a single indexed-collection query per call, consistent with the existing `GetByTenantIdAsync`.

---

### Files Reviewed

| `XiansAi.Server.Src/Shared/Services/TenantService.cs` | +10/-0 | 🟡 Medium | New duplicate check; race condition per Warning above |
| `XiansAi.Server.Src/Features/AdminApi/Services/BootstrapService.cs` | +6/-2 | 🟡 Medium | Uses new case-insensitive lookup; same check-then-act pattern, lower risk |
| `XiansAi.Server.Src/Shared/Repositories/TenantRepository.cs` | +19/-0 | 🟢 Low | New repository method; correct collation usage for read lookup |
| `XiansAi.Server.Tests/IntegrationTests/AdminApi/AdminTenantEndpointsTests.cs` | +33/-0 | 🟢 Low | New integration test covers the primary case-insensitive-duplicate scenario |

XiansAi.Server.Src/Shared/Services/TenantService.cs
```diff
@@ -637,6 +637,16 @@ public async Task<ServiceResult<TenantCreatedResult>> CreateTenant(CreateTenantR
             var validatedTenant = tenant.SanitizeAndValidate();
             _logger.LogInformation("After sanitization, CreatedBy: {CreatedBy}", LogSanitizer.Sanitize(validatedTenant.CreatedBy));
 
+            // Tenant ids are unique ignoring case: reject "MyTenant" when "mytenant" already exists.
+            // The unique index on tenant_id remains the backstop for exact-case duplicates under races.
+            var duplicateTenant = await _tenantRepository.GetByTenantIdCaseInsensitiveAsync(validatedTenant.TenantId);
```
hasith, line 642: **⚠️ WARNING — check-then-act race on case-insensitive tenant ID uniqueness**

This duplicate check only queries the database before inserting; it does not enforce uniqueness atomically. The `tenant_id_1` index defined in `mongodb-indexes.yaml` is a plain unique index with no collation, so it is case-**sensitive** by default and will not catch two concurrent `CreateTenant` calls for e.g. `MyTenant` and `mytenant`: both requests can see "no existing tenant" from `GetByTenantIdCaseInsensitiveAsync`, both proceed to `CreateAsync`, and the DB index does not reject the second write (the exact strings differ). The comment above this line ("The unique index on tenant_id remains the backstop for exact-case duplicates under races") is accurate for exact-case collisions but does not cover this PR's actual goal — case-insensitive duplicate prevention — under concurrency.

**Fix:** Add a case-insensitively-collated unique index on `tenant_id` (MongoDB supports per-index default collation, e.g. strength 2, matching the lookup's collation), or catch the MongoDB duplicate-key exception (E11000) from `CreateAsync` and convert it into the same `ServiceResult<TenantCreatedResult>.BadRequest("A tenant with this ID already exists.")` response, so the database — not just the pre-check — enforces the invariant.

PR #459, opened 2026-07-30
Feature/temporal tenant isolation

Diyanada99x:


XiansAi.Server.Src/Shared/Repositories/TenantTemporalConfigRepository.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        try
+        {
+            return await _collection.Find(x => x.TenantId == tenantId).FirstOrDefaultAsync();
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Error retrieving Temporal config for tenant {TenantId}", LogSanitizer.Sanitize(tenantId));
+            return null;
+        }
```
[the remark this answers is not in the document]
Diyanada99x: Fixed

hasith (commented):
## PR Review Report

**PR:** Feature/temporal tenant isolation (#459)
**Author:** Diyanada99x
**Files Changed:** 9 | **+307** / **-51**
**Verdict:** REQUEST CHANGES

---

### Summary
This PR adds per-tenant dedicated Temporal namespace/connection isolation (host, namespace, mTLS certificate/key stored per tenant in a new `tenant_temporal_config` collection). The design direction is reasonable and reuses the existing tenant-config fallback pattern, but the change **breaks the test build** (a constructor signature change wasn't propagated to an existing test file), stores certificate/private-key material in plaintext unlike the codebase's established encryption pattern for tenant secrets, ships a leftover debug `Console.WriteLine`, and silently drops a performance guard that prevented redundant Temporal API calls. Ships with zero new tests for ~5 new/changed classes.

---

- [x] `XiansAi.Server.Tests/UnitTests/Shared/Services/TenantServiceMetadataTests.cs:55` — The test project fails to build. `TenantService`'s constructor now requires 8 parameters (`ITenantTemporalConfigService` was inserted before `IWebhookEventPublisher`), but this test still constructs it with only 7 arguments. Verified locally: `dotnet build XiansAi.Server.Tests` fails with `CS7036: There is no argument given that corresponds to the required parameter 'metadataProtector'`.
  ```
  // Current (problematic) — line 55
  _service = new TenantService(
      _repo.Object,
      _cache.Object,
      NullLogger<TenantService>.Instance,
      _context.Object,
      _roles.Object,
      _webhooks.Object,
      _protector);

  // Fix
  private readonly Mock<ITenantTemporalConfigService> _tenantTemporalConfigService = new();
  ...
  _service = new TenantService(
      _repo.Object,
      _cache.Object,
      NullLogger<TenantService>.Instance,
      _context.Object,
      _roles.Object,
      _tenantTemporalConfigService.Object,
      _webhooks.Object,
      _protector);
  ```

- [x] `XiansAi.Server.Src/Shared/Data/Models/TenantTemporalConfig.cs:31` — The Temporal mTLS client certificate and private key (`Certificate`/`CertificateKey`) are persisted as **plaintext strings** in MongoDB, unlike the codebase's established pattern for tenant secrets: `TenantOidcConfigService` encrypts its payload with `ISecureEncryptionService` before storing (see `TenantOidcConfigService.cs:208`), and `TenantMetadataProtector` protects tenant metadata. Anyone with read access to the Mongo database (backup, DBA account, leaked connection string) can extract a tenant's Temporal mTLS private key and fully impersonate that tenant's Temporal client.
  ```
  // Current (problematic)
  [BsonElement("certificate")]
  public string? Certificate { get; set; }

  [BsonElement("certificate_key")]
  public string? CertificateKey { get; set; }

  // Fix — encrypt before persisting, mirroring TenantOidcConfigService
  // In TenantTemporalConfigService.SaveAsync, encrypt certificate/certificateKey via
  // ISecureEncryptionService before assigning to the model, and decrypt in GetForTenantAsync.
  ```

*(2 critical issues found.)*

---

### Warnings (Should Fix)

- [x] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs:105` — Leftover debug statement `Console.WriteLine($"BEFORE ENSURE***************************************: {config.FlowServerUrl}")` writes the tenant's Temporal host/URL to unstructured stdout on every new client connection, bypassing the logging pipeline. Flagged independently by all four reviewers. Remove it — the surrounding code already logs via `_logger`.

- [x] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs:166` — The refactor **dropped** the early-return guard that previously short-circuited `EnsureSearchAttributesRegisteredAsync` when a namespace's search attributes were already known to be registered (confirmed via diff: `if (_searchAttributesRegistered.TryGetValue(namespaceName, out var registered) && registered) return;` was removed). Every tenant connection now unconditionally calls `ListSearchAttributesAsync`/potentially `AddSearchAttributesAsync` against Temporal, even for namespaces already registered — a real regression, not just a style nit.
  ```
  // Fix — restore at the top of EnsureSearchAttributesRegisteredAsync
  private async Task EnsureSearchAttributesRegisteredAsync(ITemporalClient client, string namespaceName)
  {
      if (_searchAttributesRegistered.TryGetValue(namespaceName, out var registered) && registered)
          return;

      try
      { ... }
  ```

- [x] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs:127` — `GetTemporalConfigAsync` decides whether to use the tenant-dedicated Temporal connection purely by checking whether a `TenantTemporalConfig` document exists — it never checks `Tenant.UseSpecificTemporalNamespace`. Combined with the fact that `UpdateTenantRequest` has no `TemporalHost`/`TemporalCertificate`/etc. fields (confirmed: those only exist on the create-request model) and there is no `DeleteAsync`/revoke method on `ITenantTemporalConfigRepository`/`ITenantTemporalConfigService`, flipping `UseSpecificTemporalNamespace` back to `false` via `UpdateTenant` does **not** stop the server from using the previously stored (and un-rotatable) tenant-specific host/certificate/key. The flag is effectively decorative for the read path, and there is no admin-facing way to rotate or revoke a compromised tenant certificate after creation.

- [x] `XiansAi.Server.Src/Shared/Services/CertificateService.cs:49` — `GetFlowServerSettings()` now performs an unconditional MongoDB round trip via `GetForTenantAsync` on **every** call, even for the majority of tenants that never opted into `UseSpecificTemporalNamespace`, with no caching. This is a per-request settings endpoint; consider caching the (mostly-null) lookup result or gating it behind the tenant's flag.

- [x] `XiansAi.Server.Src/Shared/Services/CertificateService.cs:52,62` — These two log statements use raw string interpolation (`$"..."`) instead of the `LogSanitizer`/structured-logging pattern used elsewhere in this same file (e.g. lines 105-107, 151-155), which is inconsistent and risks unsanitized values reaching logs.

*(5 warnings found.)*

---

- [x] `XiansAi.Server.Src/Shared/Repositories/TenantTemporalConfigRepository.cs:29` — `GetByTenantIdAsync` filters `tenant_temporal_config` by `tenant_id` with no index created anywhere in the PR, unlike `AppIntegrationRepository`'s explicit `idx_tenant_id` index. Lookups will degrade to collection scans as tenant count grows.

- [ ] `XiansAi.Server.Src/Shared/Services/TenantService.cs:679` — `_tenantTemporalConfigService.SaveAsync(...)` persists the tenant's Temporal config (including cert/key) **before** `_tenantRepository.CreateAsync(validatedTenant)` runs. If tenant creation subsequently fails (e.g. the duplicate-key race handled at the catch block), an orphaned `TenantTemporalConfig` document remains for a tenant that was never created. Consider ordering the two writes so the config is only persisted after the tenant record commits, or cleaning up on failure.

- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs:22,33,38` — Commented-out `ITenantCacheService` field/parameter/assignment left in the file adds noise; remove if unused or track as a follow-up if intentional.

- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs:53` — `ITemporalClient` instances are cached indefinitely in `_clients` keyed by `tenantId` with no invalidation when `TenantTemporalConfigService.SaveAsync` updates a tenant's config. Rotating a tenant's certificate won't take effect until process restart.

- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs:68` — The single global `_connectionSemaphore` (capacity 1) now also guards the new async Mongo lookup plus the TLS handshake for every tenant's first connection; a slow Mongo call or handshake for one tenant serializes first-time connection setup for all other tenants.

---

### Review Details

#### Code Quality
Overall structure follows existing patterns (fallback chain: tenant-specific Mongo config → per-tenant `IConfiguration` section → root default). Main issues: a leftover debug `Console.WriteLine`, a silently-dropped performance guard in `EnsureSearchAttributesRegisteredAsync`, unsanitized log interpolation in `CertificateService`, and dead/commented-out code for an unused `ITenantCacheService` dependency. Input validation in `TenantTemporalConfigService.SaveAsync` (host/namespace required, cert+key must be provided together) is a good addition, and error handling in the new repository consistently logs via `LogSanitizer` before rethrowing.

#### Security
Two real issues: (1) Temporal mTLS certificate/private key stored in plaintext in Mongo, inconsistent with this codebase's own encryption pattern for tenant secrets (`TenantOidcConfigService`); (2) the `UseSpecificTemporalNamespace` toggle is non-authoritative — there's no code path to disable/rotate a tenant's stored dedicated connection once created. No SQL/NoSQL injection, hardcoded secrets, or authorization gaps were found; tenant create/update endpoints are correctly gated by sysadmin authorization, and the `GetByTenantIdAsync` rethrow change does not leak exception details to callers (caught and converted to a generic 500 upstream).

#### Test Coverage
**0% coverage** of new/changed functionality: `TenantTemporalConfig` (new model), `TenantTemporalConfigRepository` (new, 2 methods), `TenantTemporalConfigService` (new, 2 methods), the new temporal-config branch in `TenantService.CreateTenant`, the async rewrite of `CertificateService.GetFlowServerSettings`, and the `TemporalClientService` refactor (including the dropped search-attributes cache) all ship with no tests. Worse, this PR **breaks the existing build**: `TenantServiceMetadataTests.cs` doesn't compile against the new `TenantService` constructor signature (verified via `dotnet build`).

#### Performance
No N+1/O(n²) patterns, but three real concerns: the dropped early-return cache check causes redundant Temporal API calls per connection; `CertificateService.GetFlowServerSettings` now does an unconditional, uncached Mongo round-trip on a per-request settings endpoint; and the new `tenant_temporal_config` collection has no index on `tenant_id`.

---

### Files Reviewed

| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs` | +82/-46 | 🔴 High | Debug leftover, dropped perf guard, cert/connection caching |
| `XiansAi.Server.Src/Shared/Data/Models/TenantTemporalConfig.cs` | +44/-0 | 🔴 High | New model — plaintext cert/key storage |
| `XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs` | +73/-0 | 🔴 High | New service — writes plaintext secrets |
| `XiansAi.Server.Src/Shared/Services/TenantService.cs` | +44/-3 | 🟠 Medium | Constructor signature change (breaks test build), ordering of writes |
| `XiansAi.Server.Src/Shared/Services/CertificateService.cs` | +23/-1 | 🟠 Medium | New Mongo call on hot path, log sanitization |
| `XiansAi.Server.Src/Shared/Repositories/TenantTemporalConfigRepository.cs` | +51/-0 | 🟡 Low-Medium | New repository, no index |
| `XiansAi.Server.Src/Shared/Data/Models/Tenant.cs` | +7/-3 | 🟢 Low | New flag + serialization |
| `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs` | +2/-0 | 🟢 Low | DI registration |
| `XiansAi.Server.Src/Features/AgentApi/Endpoints/SettingsEndpoints.cs` | +4/-1 | 🟢 Low | Await added for async call |

XiansAi.Server.Src/Shared/Data/Models/TenantTemporalConfig.cs
```diff
@@ -0,0 +1,44 @@
...
+    public required string Host { get; set; }
+
+    [BsonElement("namespace")]
+    public required string Namespace { get; set; }
+
+    [BsonElement("certificate")]
+    public string? Certificate { get; set; }
+
+    [BsonElement("certificate_key")]
+    public string? CertificateKey { get; set; }
```
hasith, line 31: **CRITICAL:** The Temporal mTLS client certificate and private key (`Certificate`/`CertificateKey`) are persisted as **plaintext strings** in the `tenant_temporal_config` MongoDB collection, with no encryption at rest.

This is inconsistent with this codebase's established pattern for tenant secrets: `TenantOidcConfigService` encrypts its payload via `ISecureEncryptionService` before storing (`_encryption.Encrypt(jsonConfig, _uniqueSecret)`), and `TenantMetadataProtector` protects tenant metadata secrets the same way.

Anyone with read access to the Mongo database (backup dump, DBA account, leaked connection string) can extract a tenant's Temporal mTLS private key and fully impersonate that tenant's Temporal client.

**Fix:** Inject `ISecureEncryptionService` into `TenantTemporalConfigService`, encrypt `certificate`/`certificateKey` before persisting (mirroring `TenantOidcConfigService`), and decrypt on read in `GetForTenantAsync`/`CertificateService.GetFlowServerSettings`.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs
```diff
@@ -58,63 +65,82 @@ public async Task<ITemporalClient> GetClientAsync(string tenantId)
...
-                return client;
-            }
-            catch (Exception ex)
-            {
-                _logger.LogError(ex, "Failed to connect to Temporal server for tenant {TenantId} at {Url}. Error: {ErrorMessage}",
-                    tenantId, config.FlowServerUrl, ex.Message);
-                _clients.TryRemove(tenantId, out _);
-                throw;
-            }
+                    Console.WriteLine($"BEFORE ENSURE***************************************: {config.FlowServerUrl}");
```
hasith: **Warning:** Leftover debug statement writes the tenant's Temporal host/URL to unstructured stdout on every new client connection, bypassing the logging pipeline. The surrounding code already logs via `_logger`.

**Fix:** Remove the line.

```suggestion
```

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs
```diff
@@ -137,8 +163,6 @@ private TemporalConfig GetTemporalConfig(string tenantId)
     /// </summary>
     private async Task EnsureSearchAttributesRegisteredAsync(ITemporalClient client, string namespaceName)
     {
-        if (_searchAttributesRegistered.TryGetValue(namespaceName, out var registered) && registered)
-            return;
```
hasith, line 172: **Warning:** This refactor dropped the early-return guard that previously short-circuited `EnsureSearchAttributesRegisteredAsync` when a namespace's search attributes were already known to be registered. Confirmed via diff — this line was removed:
```csharp
if (_searchAttributesRegistered.TryGetValue(namespaceName, out var registered) && registered)
    return;
```
Every tenant connection now unconditionally calls `ListSearchAttributesAsync` (and potentially `AddSearchAttributesAsync`) against Temporal, even for namespaces already registered. This is a real performance regression, not a style nit.

**Fix:** Restore the guard at the top of the method.

```suggestion
        if (_searchAttributesRegistered.TryGetValue(namespaceName, out var registered) && registered)
            return;

```

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs
```diff
@@ -58,63 +65,82 @@ public async Task<ITemporalClient> GetClientAsync(string tenantId)
...
-    private TemporalConfig GetTemporalConfig(string tenantId)
+    private async Task<TemporalConfig> GetTemporalConfigAsync(string tenantId)
     {
         if (string.IsNullOrEmpty(tenantId))
             throw new InvalidOperationException("TenantId is required");
 
-        // First try to get tenant-specific temporal config
+        // A tenant that opted into "use a specific Temporal namespace" has its connection
+        // details persisted in Mongo - prefer that over static configuration.
+        var tenantConnection = await _tenantTemporalConfigService.GetForTenantAsync(tenantId);
```
hasith: **Warning:** `GetTemporalConfigAsync` decides whether to use the tenant-dedicated Temporal connection purely by checking whether a `TenantTemporalConfig` document exists (`if (tenantConnection != null)`) — it never checks `Tenant.UseSpecificTemporalNamespace`.

Combined with two other gaps confirmed in this PR: `UpdateTenantRequest` has no `TemporalHost`/`TemporalCertificate`/etc. fields (those only exist on the create-request model), and there is no `DeleteAsync`/revoke method on `ITenantTemporalConfigRepository`/`ITenantTemporalConfigService` — flipping `UseSpecificTemporalNamespace` back to `false` via `UpdateTenant` does **not** stop the server from using the previously stored (and un-rotatable) tenant-specific host/certificate/key. The toggle is effectively decorative for the read path, and there's no admin-facing way to rotate or revoke a compromised tenant certificate after creation.

**Fix:** Gate this lookup on `tenant.UseSpecificTemporalNamespace` (requires loading/passing the flag), and add a delete/disable method so an admin can actually revoke a dedicated Temporal connection.

XiansAi.Server.Src/Shared/Services/CertificateService.cs
```diff
@@ -23,23 +23,42 @@ public class CertificateService
...
+        _tenantTemporalConfigService = tenantTemporalConfigService;
     }
 
-    public FlowServerSettings GetFlowServerSettings()
+    public async Task<FlowServerSettings> GetFlowServerSettings()
     {
+        // A tenant that opted into "use a specific Temporal namespace" has its connection
+        // details persisted in Mongo (tenant_temporal_config) - prefer that over static
+        // configuration, mirroring TemporalClientService.GetTemporalConfigAsync.
+        var tenantConnection = await _tenantTemporalConfigService.GetForTenantAsync(_tenantContext.TenantId);
```
hasith: **Warning:** `GetFlowServerSettings()` now performs an unconditional MongoDB round trip via `GetForTenantAsync` on every call, even for the majority of tenants that never opted into `UseSpecificTemporalNamespace`, with no caching of the (mostly-null) result. This is a per-request settings endpoint.

**Fix:** Short-circuit using the tenant's already-loaded `UseSpecificTemporalNamespace` flag, or cache `GetForTenantAsync` results (e.g. `IMemoryCache` keyed by tenant id, invalidated on save) before hitting Mongo on this hot path.

XiansAi.Server.Src/Shared/Services/CertificateService.cs
```diff
@@ -23,23 +23,42 @@ public class CertificateService
...
+        if (tenantConnection != null)
+        {
+            _logger.LogInformation($"GetFlowServerSettings for Tenant:{_tenantContext.TenantId} FlowServerUrl:{tenantConnection.Host} FlowServerNamespace:{tenantConnection.Namespace}");
```
hasith: **Warning:** This log statement (and the similar one at line 62) uses raw string interpolation (`$"..."`) instead of the `LogSanitizer`/structured-logging pattern used elsewhere in this same file (e.g. lines 105-107, 151-155). This is inconsistent with the rest of the codebase and risks unsanitized values reaching logs.

**Fix:** Convert to structured logging with `LogSanitizer.Sanitize(...)` on the interpolated values, matching the pattern used elsewhere in `CertificateService`.

XiansAi.Server.Src/Shared/Repositories/TenantTemporalConfigRepository.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        _collection = mongoDbClientService.GetCollection<TenantTemporalConfig>("tenant_temporal_config");
+        _logger = logger;
+    }
+
+    public async Task<TenantTemporalConfig?> GetByTenantIdAsync(string tenantId)
+    {
+        try
+        {
+            return await _collection.Find(x => x.TenantId == tenantId).FirstOrDefaultAsync();
```
hasith, line 29: **Suggestion:** `GetByTenantIdAsync` filters the new `tenant_temporal_config` collection by `tenant_id` with no index defined anywhere in this PR, unlike `AppIntegrationRepository` which explicitly creates an `idx_tenant_id` index. Lookups will degrade to collection scans as tenant count grows.

**Fix:** Create an index on `tenant_id` for the `tenant_temporal_config` collection at startup, mirroring `AppIntegrationRepository`'s `CreateIndexOptions { Name = "idx_tenant_id" }` pattern.

XiansAi.Server.Src/Shared/Services/TenantService.cs
```diff
@@ -649,6 +658,32 @@ public async Task<ServiceResult<TenantCreatedResult>> CreateTenant(CreateTenantR
...
+                var hasCertificate = !string.IsNullOrWhiteSpace(request.TemporalCertificate);
+                var hasCertificateKey = !string.IsNullOrWhiteSpace(request.TemporalCertificateKey);
+                if (hasCertificate != hasCertificateKey)
+                {
+                    return ServiceResult<TenantCreatedResult>.BadRequest(
+                        "Temporal certificate and certificate key must be provided together.");
+                }
+
+                _logger.LogInformation("Saving dedicated Temporal connection for tenant {TenantId}", LogSanitizer.Sanitize(request.TenantId));
+                await _tenantTemporalConfigService.SaveAsync(
```
hasith, line 679: **Suggestion:** `_tenantTemporalConfigService.SaveAsync(...)` persists the tenant's Temporal config (including cert/key) *before* `_tenantRepository.CreateAsync(validatedTenant)` runs below. If tenant creation subsequently fails (e.g. the duplicate-key race handled in the catch block further down), an orphaned `TenantTemporalConfig` document remains for a tenant that was never actually created.

**Fix:** Reorder so the Temporal config is only persisted after `CreateAsync` succeeds, or add cleanup on failure.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs
```diff
@@ -2,22 +2,25 @@
...
     ITemporalClient GetClient(string tenantId);
-
 }
 
 public class TemporalClientService : ITemporalClientService, IDisposable, IAsyncDisposable
 {
     private readonly ConcurrentDictionary<string, ITemporalClient> _clients = new();
     private readonly ConcurrentDictionary<string, CloudService> _serviceClients = new();
     private readonly ConcurrentDictionary<string, bool> _searchAttributesRegistered = new();
+    // private readonly ITenantCacheService _tenantCacheService;
```
hasith: **Suggestion:** Commented-out `ITenantCacheService` field/constructor-parameter/assignment (here and at lines 33 and 38) adds noise and risks confusion about which dependency is actually in use.

**Fix:** Remove the commented-out lines, or track as a follow-up if this is intentionally staged for a future change.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs
```diff
@@ -58,63 +65,82 @@ public async Task<ITemporalClient> GetClientAsync(string tenantId)
                 return existingClient;
             }
 
-            var config = GetTemporalConfig(tenantId);
-            
-            var options = new TemporalClientConnectOptions(new(config.FlowServerUrl))
+            var config = await GetTemporalConfigAsync(tenantId);
```
hasith, line 68: **Suggestion:** The single global `_connectionSemaphore` (capacity 1) now also guards the new async Mongo lookup (`GetTemporalConfigAsync`) plus the TLS handshake for every tenant's first connection, so a slow/unavailable Mongo call or handshake for one tenant serializes first-time connection setup for all other tenants platform-wide.

**Fix:** Use a per-tenant lock (e.g. `ConcurrentDictionary<string, SemaphoreSlim>` keyed by tenantId) instead of one process-wide semaphore.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs
```diff
@@ -2,22 +2,25 @@
...
 public interface ITemporalClientService
 {
     Task<ITemporalClient> GetClientAsync(string tenantId);
     ITemporalClient GetClient(string tenantId);
-
 }
 
 public class TemporalClientService : ITemporalClientService, IDisposable, IAsyncDisposable
 {
     private readonly ConcurrentDictionary<string, ITemporalClient> _clients = new();
```
hasith, line 19: **Suggestion:** `ITemporalClient` instances are cached indefinitely in `_clients` keyed by `tenantId`, with no invalidation when `TenantTemporalConfigService.SaveAsync` updates a tenant's stored host/namespace/certificate. Rotating a tenant's certificate (e.g. after suspected compromise) won't take effect until process restart.

**Fix:** Invalidate/evict the corresponding `_clients` entry when a tenant's temporal config is updated (e.g. via an event/callback from `TenantTemporalConfigService`).

hasith (commented):
## PR Review Report

**PR:** Feature/temporal tenant isolation (#459)
**Author:** Diyanada99x
**Files Changed:** 10 | **+307** / **-51**
**Verdict:** REQUEST CHANGES

---

### Re-review delta

Reviewed 2 new commit(s) since the last review (`a0d3062`..`d04943b`).
- ✅ Fixed: 1 previously-flagged issue resolved (leftover `Console.Write` debug statement in `TemporalClientService.cs` was removed)
- ⏳ Still open: 10 carried-over issue(s), including the CRITICAL plaintext secret storage finding
- 🆕 New: 5 issue(s) identified in this pass

---

### Existing review threads

- ✅ Appears addressed: 1 open thread (debug `Console.Write` statement — removed in commit `f2fed03`)
- ⏳ Still open: 10 open thread(s) — no reply (avoid spam)
- 🔇 Duplicates avoided: 4 finding(s) not re-posted as new inline threads (already covered by open threads: `CertificateService.cs` unstructured logging at lines 52/62, `TenantService.cs:679` save-before-create ordering, `TemporalClientService.cs:22` commented-out dead code). Note: the `TenantService.cs:680` unsanitized-tenant-ID finding below was mechanically folded into the adjacent open thread at line 679 by proximity dedup even though it is a distinct concern (isolation-breaking bug vs. orphan-on-failure risk) — flagged explicitly in Warnings below since it did not get its own inline thread.

---

### Summary

This PR adds per-tenant Temporal isolation (dedicated host/namespace/mTLS cert per tenant) via a new `TenantTemporalConfig` model, repository, and service, wired into `TenantService`, `CertificateService`, and `TemporalClientService`. Since the last review, the leftover debug `Console.Write` was removed and a test mock was added to keep the build green, but the standing CRITICAL issue (tenant Temporal mTLS certificates/keys persisted as plaintext, unlike this codebase's other tenant-secret paths) remains unaddressed, and this pass surfaces a new correctness bug where the tenant's dedicated Temporal config is saved under the *unsanitized* tenant ID, which can silently break the isolation this PR is meant to deliver. The new code also ships with zero unit/integration test coverage.

---

No **new** critical issues introduced since the last review. The following CRITICAL issue from the prior review remains open and unaddressed at `HEAD` (see existing open thread, not re-posted here to avoid duplication):

- [ ] `XiansAi.Server.Src/Shared/Data/Models/TenantTemporalConfig.cs:31` — Temporal mTLS `Certificate`/`CertificateKey` are persisted as plaintext in MongoDB with no encryption at rest, inconsistent with `TenantOidcConfigService`'s use of `ISecureEncryptionService`. Anyone with DB/backup read access can extract a tenant's Temporal mTLS private key.

---

### Warnings (Should Fix)

- [ ] `XiansAi.Server.Src/Shared/Services/TenantService.cs:680` — `SaveAsync` for the tenant's dedicated Temporal config is called with the raw, unsanitized `request.TenantId` instead of `validatedTenant.TenantId`. If sanitization changes the ID at all, the config is stored under an ID that later lookups will never match, so the tenant silently falls back to the shared Temporal namespace with no error — defeating this PR's isolation goal for that tenant.
  ```csharp
  // Current (problematic)
  _logger.LogInformation("Saving dedicated Temporal connection for tenant {TenantId}", LogSanitizer.Sanitize(request.TenantId));
  await _tenantTemporalConfigService.SaveAsync(
      request.TenantId,
      ...

  // Fix
  _logger.LogInformation("Saving dedicated Temporal connection for tenant {TenantId}", LogSanitizer.Sanitize(validatedTenant.TenantId));
  await _tenantTemporalConfigService.SaveAsync(
      validatedTenant.TenantId,
      ...
  ```
- [ ] `XiansAi.Server.Src/Shared/Services/TenantService.cs:765` — `UpdateTenant` allows flipping `UseSpecificTemporalNamespace` back to `false` but never clears/revokes the tenant's stored `TenantTemporalConfig` or invalidates its cached `ITemporalClient`, so the old dedicated connection (and its cert/key) stays live.
- [ ] `XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs:13` — The new tenant-secret-handling classes (`TenantTemporalConfig`, `TenantTemporalConfigRepository`, `TenantTemporalConfigService`) ship with zero unit/integration tests; only an existing test file was touched to inject a mock dependency.

*(10 additional carried-over warnings/suggestions from the prior review remain open — see existing PR review threads; not duplicated here.)*

---

- [ ] `XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs:22` — `SaveAsync` doesn't reject a blank/whitespace `createdBy`, allowing an empty audit-trail owner to be persisted.
- [ ] `XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs:26` — No validation that `host` is a well-formed, non-internal endpoint before it's used to open a Temporal connection. Low risk today (route is `SysAdmin`-only) but worth hardening as defense-in-depth (SSRF) before any self-service onboarding is added.
- [ ] `XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs:33` — `SaveAsync` does a read (`GetByTenantIdAsync`) then a write (`UpsertAsync`) instead of a single atomic upsert, costing an extra Mongo round trip and opening a small read/write race window.

---

### Review Details

#### Code Quality
Overall the new `TenantTemporalConfig`/`TenantTemporalConfigRepository`/`TenantTemporalConfigService` trio follows the codebase's existing repository/service layering, and the `Tenant` model correctly threads the new `UseSpecificTemporalNamespace` flag through `Copy()`/`SanitizeAndReturn()`. The unsanitized-tenant-ID bug at `TenantService.cs:680` and the missing config cleanup on flag-disable (`:765`) are the notable gaps; minor validation and dead-code cleanup items are already tracked in prior-review threads.

#### Security
The standing CRITICAL finding (plaintext mTLS certificate/key storage) is unaddressed. All other previously-flagged security/hardening issues (missing revocation on flag toggle, unconditional Mongo round-trip, unsanitized logging, missing index, semaphore serialization, no cache invalidation) remain open. This pass adds one new correctness/isolation bug (unsanitized tenant ID used when persisting dedicated Temporal config) and one low-severity SSRF hardening suggestion. The previously-flagged leftover debug statement has been removed.

#### Test Coverage
0% of the new/modified Temporal-isolation code paths have dedicated tests: `TenantTemporalConfigRepository`, `TenantTemporalConfigService`, the new `CreateTenant`/`UpdateTenant` branches in `TenantService`, the refactored `CertificateService.GetFlowServerSettings()`, and `TemporalClientService.GetTemporalConfigAsync()` are all untested. The one touched test file (`TenantServiceMetadataTests.cs`) was updated only to keep the existing suite compiling (new mock dependency), not to exercise the new feature.

#### Performance
The read-then-write pattern in `TenantTemporalConfigService.SaveAsync` is a minor, non-hot-path inefficiency. The more significant performance concerns (unconditional per-request Mongo lookup in `CertificateService`, missing `tenant_id` index, and the global connection semaphore now also guarding the Mongo lookup) were flagged in the prior review and remain open.

---

### Files Reviewed

| `XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs` | +73/-0 | 🔴 High | New service; secret handling, zero tests |
| `XiansAi.Server.Src/Shared/Services/TenantService.cs` | +44/-4 (approx) | 🔴 High | Unsanitized tenant ID bug; stale config on disable |
| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs` | +58/-50 (approx) | 🔴 High | Client caching, connection setup; carried-over findings |
| `XiansAi.Server.Src/Shared/Services/CertificateService.cs` | +19/-4 (approx) | 🟡 Medium | Carried-over logging/perf findings |
| `XiansAi.Server.Src/Shared/Repositories/TenantTemporalConfigRepository.cs` | +51/-0 | 🟡 Medium | New repository; missing index (carried-over), zero tests |
| `XiansAi.Server.Src/Shared/Data/Models/TenantTemporalConfig.cs` | +44/-0 | 🔴 High | Plaintext secret storage (carried-over CRITICAL) |
| `XiansAi.Server.Src/Shared/Data/Models/Tenant.cs` | +7/-0 | 🟢 Low | New flag property, threaded through Copy/Sanitize |
| `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs` | +2/-0 | 🟢 Low | DI registration |
| `XiansAi.Server.Src/Features/AgentApi/Endpoints/SettingsEndpoints.cs` | +4/-2 | 🟢 Low | Async signature update |
| `XiansAi.Server.Tests/UnitTests/Shared/Services/TenantServiceMetadataTests.cs` | +2/-0 | 🟢 Low | Mock injection only, no new test cases |

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs
hasith: ✅ Resolved as of `d04943b1510837d215811fa9c4067b788ae9af5b`. This finding no longer reproduces against the current head.

XiansAi.Server.Src/Shared/Services/TenantService.cs
```diff
@@ -727,6 +762,9 @@ public async Task<ServiceResult<Tenant>> UpdateTenant(string id, UpdateTenantReq
             if (request.Timezone != null)
                 existingTenant.Timezone = request.Timezone;
 
+            if (request.UseSpecificTemporalNamespace.HasValue)
```
hasith, line 765: **Warning:** `UpdateTenant` lets an admin flip `UseSpecificTemporalNamespace` back to `false`, but nothing here deletes or clears the corresponding `TenantTemporalConfig` document, and `TemporalClientService` caches `ITemporalClient` instances indefinitely in `_clients` keyed by tenant id (see the already-open thread on that file). The net effect: disabling dedicated Temporal isolation for a tenant does not actually revoke/rotate its stored connection details or invalidate its cached client.

**Fix:** When `UseSpecificTemporalNamespace` transitions from `true` to `false` (or the underlying host/namespace/cert is otherwise rotated), also remove the tenant's `TenantTemporalConfig` entry and invalidate/dispose its cached client in `TemporalClientService`.

XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Shared.Services;
+
+public interface ITenantTemporalConfigService
+{
+    Task<TenantTemporalConfig> SaveAsync(string tenantId, string host, string @namespace, string? certificate, string? certificateKey, string createdBy);
+    Task<TenantTemporalConfig?> GetForTenantAsync(string tenantId);
+}
+
+public class TenantTemporalConfigService : ITenantTemporalConfigService
```
hasith, line 15: **Warning:** This PR ships three new production classes that handle tenant Temporal connection details and mTLS secrets — `TenantTemporalConfig`, `TenantTemporalConfigRepository`, and `TenantTemporalConfigService` — with zero accompanying unit or integration tests. The only test file touched (`TenantServiceMetadataTests.cs`) was updated solely to inject the new mock dependency; no test exercises `SaveAsync`'s validation rules (required fields, certificate/key pairing), `GetForTenantAsync`, or the repository's `GetByTenantIdAsync`/`UpsertAsync`, nor the new `UseSpecificTemporalNamespace` branch in `TenantService.CreateTenant`/`UpdateTenant`.

**Fix:** Add unit tests for `TenantTemporalConfigService.SaveAsync` (create vs. update path, validation errors, cert/key parity rule) and `GetForTenantAsync`, plus repository tests for `TenantTemporalConfigRepository` following the existing `MongoDbFixture`-based patterns, and service-level tests for the new `CreateTenant`/`UpdateTenant` branches.

XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    private readonly ITenantTemporalConfigRepository _repository;
+
+    public TenantTemporalConfigService(ITenantTemporalConfigRepository repository)
+    {
+        _repository = repository ?? throw new ArgumentNullException(nameof(repository));
+    }
+
+    public async Task<TenantTemporalConfig> SaveAsync(string tenantId, string host, string @namespace, string? certificate, string? certificateKey, string createdBy)
```
hasith, line 45: **Suggestion:** `SaveAsync` validates `host`/`@namespace`/certificate-pairing but not `createdBy`, so a blank or whitespace-only `createdBy` can be silently persisted as the audit trail owner for a tenant's Temporal connection config.

**Fix:** Add `if (string.IsNullOrWhiteSpace(createdBy)) throw new ArgumentException("createdBy is required", nameof(createdBy));` alongside the other guards.

XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        if (string.IsNullOrWhiteSpace(tenantId))
+            throw new ArgumentException("tenantId is required", nameof(tenantId));
+        if (string.IsNullOrWhiteSpace(host))
```
hasith, line 49: **Suggestion:** `SaveAsync` never validates that `host` is a well-formed, non-internal endpoint before it is later fed straight into `TemporalClient.ConnectAsync` options in `TemporalClientService`. Today the blast radius is limited because both `CreateTenant` routes that reach this path require the `SysAdmin` role, but if tenant self-service onboarding is ever added without revisiting this path it becomes an SSRF primitive.

**Fix:** Add basic scheme/host validation as defense-in-depth (e.g. reject `localhost`/link-local/metadata addresses) before persisting.

XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs
```diff
@@ -0,0 +1,73 @@
...
+            throw new ArgumentException("host is required", nameof(host));
+        if (string.IsNullOrWhiteSpace(@namespace))
+            throw new ArgumentException("namespace is required", nameof(@namespace));
+        if (string.IsNullOrWhiteSpace(certificate) != string.IsNullOrWhiteSpace(certificateKey))
+            throw new ArgumentException("certificate and certificateKey must be provided together");
+
+        var existing = await _repository.GetByTenantIdAsync(tenantId);
```
hasith, line 59: **Suggestion:** `SaveAsync` does a `GetByTenantIdAsync` read followed by an `UpsertAsync` write (lines 33 and 62), turning a logically-atomic upsert into two sequential MongoDB round trips and opening a race window where two concurrent `SaveAsync` calls for the same tenant can interleave between the read and the write (last-writer-wins on `CreatedAt`/`CreatedBy`).

**Fix:** Replace the read-then-write with a single `UpdateOneAsync(filter, update, new UpdateOptions { IsUpsert = true })` using `$setOnInsert` for `Id`/`CreatedAt`/`CreatedBy` and `$set` for the mutable fields.

hasith (commented):
## PR Review Report

**PR:** Feature/temporal tenant isolation (#459)
**Author:** Diyanada Gunawrdena
**Files Changed:** 10 | **+350** / **-49**
**Verdict:** REQUEST CHANGES

---

### Re-review delta

Reviewed 2 new commit(s) since the last review (`d04943b`..`fdb0d2d`).
- ✅ Fixed: 0 previously-flagged issue(s) confirmed resolved by the mechanical reconcile gate this run (note: the code was independently re-read and the `EnsureSearchAttributesRegisteredAsync` early-return guard *is* present and functioning at `TemporalClientService.cs:164-167` as of commit `fdb0d2d` — this looks fixed on manual inspection, but the automated fid-reconciliation gate did not confirm it this pass, so the existing thread is left open rather than auto-resolved)
- ⏳ Still open: 15 carried-over issue(s)
- 🆕 New: 4 issue(s) newly identified since the last review

---

### Existing review threads

- ✅ Appears addressed: 0 open thread(s) (see note above — not auto-resolved this run)
- ⏳ Still open: 15 open thread(s) — no reply (avoid spam)
- 🔇 Duplicates avoided: 0 — this run's automated line-proximity dedup initially flagged 3 of the 4 findings below as possible duplicates of nearby open threads (line±5 in the same file), but each was manually verified to raise a **materially different concern** from the nearby thread (see "Note" in each finding) and is posted as a genuinely new finding rather than suppressed.

---

### Summary

This PR adds per-tenant Temporal namespace isolation with encrypted mTLS certificate/key storage — a solid feature with sound crypto (AES-256-GCM via the existing `ISecureEncryptionService`, matching the established `TenantOidcConfigService` pattern). The latest commits fix the previously-flagged missing early-return guard in `EnsureSearchAttributesRegisteredAsync`. However, this pass surfaced a **new critical bug**: `TenantService.CreateTenant` saves the tenant's `TenantTemporalConfig` keyed by the **raw, unsanitized** `request.TenantId` while the `Tenant` document itself is saved under the **sanitized** `validatedTenant.TenantId` — if the two diverge, the dedicated Temporal config becomes permanently unreachable and the tenant silently falls back to the shared namespace, defeating the isolation feature. A number of substantial design/test gaps remain open from the prior review round (plaintext-at-rest history, non-atomic upsert, missing index, no cache invalidation on config rotation, zero test coverage) and are not repeated here.

---

- [ ] `XiansAi.Server.Src/Shared/Services/TenantService.cs:680` — `TenantTemporalConfigService.SaveAsync` is called with the **unsanitized** `request.TenantId`, while the `Tenant` document is persisted under the **sanitized** `validatedTenant.TenantId` (via `SanitizeAndValidate()`). If `request.TenantId` contains characters stripped during sanitization, the two IDs diverge: the encrypted mTLS cert/key config is saved under an ID that later lookups (`GetByTenantIdAsync` keyed off `_tenantContext.TenantId`, the canonical/sanitized ID) will never match. The tenant silently falls back to the shared/default namespace and an orphaned document containing an encrypted private key is left with no reachable owner.

  ```csharp
  // Current (problematic)
  await _tenantTemporalConfigService.SaveAsync(
      request.TenantId,
      request.TemporalHost,
      request.TemporalNamespace,
      hasCertificate ? request.TemporalCertificate : null,
      hasCertificateKey ? request.TemporalCertificateKey : null,
      finalCreatedBy);

  // Fix — use the already-sanitized ID
  await _tenantTemporalConfigService.SaveAsync(
      validatedTenant.TenantId,
      request.TemporalHost,
      request.TemporalNamespace,
      hasCertificate ? request.TemporalCertificate : null,
      hasCertificateKey ? request.TemporalCertificateKey : null,
      finalCreatedBy);
  ```

---

### Warnings (Should Fix)

- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs:164` — `_searchAttributesRegistered` (declared line 21) is a process-wide cache keyed **only by namespace name**. Before this PR, namespace strings came from curated static config; now any tenant admin can supply an arbitrary `Namespace` via `TenantTemporalConfigService.SaveAsync`. If two tenants (or a tenant and the default cluster) pick the same namespace string against **different** Temporal hosts, the first tenant's cached registration entry causes the second tenant's `EnsureSearchAttributesRegisteredAsync` call to short-circuit and skip verifying/registering attributes against its own, different cluster — silent cache-key collision that gets worse as tenant count grows.
  **Fix:** Key the cache by a composite of host+namespace (or tenantId+namespace) instead of namespace alone.

- [ ] `XiansAi.Server.Tests/UnitTests/Shared/Services/TenantServiceMetadataTests.cs:30,62` — The new `Mock<ITenantTemporalConfigService> _temporalConfigs` is injected into `TenantService`'s constructor but never configured with `.Setup(...)` return values anywhere in the test class. Any test path where `UseSpecificTemporalNamespace` is true will silently invoke the unconfigured mock (returning `default`/`null`), masking behavioral bugs in the `CreateTenant`/`UpdateTenant` temporal-config branches rather than exercising them.
  **Fix:** Add explicit `.Setup()` calls for `_temporalConfigs` covering the temporal-namespace branches, or add dedicated tests for those branches in a new test class.

---

- [ ] `XiansAi.Server.Src/Shared/Repositories/TenantTemporalConfigRepository.cs:34` — `GetByTenantIdAsync` rethrows on any Mongo error, unlike the sibling `TenantOidcConfigRepository.GetByTenantIdAsync` (same shape), which returns `null` on error. Not an information-disclosure risk (the global exception handler returns a generic 500), but it is a behavioral regression from the codebase's own convention: a transient Mongo hiccup now hard-fails every `GetFlowServerSettings()`/`GetClientAsync()` call for a tenant with dedicated Temporal config, instead of gracefully falling back to the default/static configuration on this availability-sensitive path. Left as author judgment since it changes failure-mode semantics.

---

### Review Details

#### Code Quality
Encryption usage is correct and consistent with the established `TenantOidcConfigService` pattern; new services (`TenantTemporalConfigService`, `TenantTemporalConfigRepository`) cleanly separate business logic from data access with proper constructor null-checks. Most other quality concerns identified this pass (missing `createdBy` validation, non-atomic read-then-write in `SaveAsync`, commented-out `ITenantCacheService` dead code, inconsistent raw-string log interpolation in `CertificateService`, tenant-creation orphan ordering) duplicate findings already tracked on open plugin threads from the prior review and are not repeated here.

#### Security
Certificate/key encryption at rest (AES-256-GCM, random 12-byte nonce, PBKDF2-SHA256 200k-iteration key derivation) is implemented correctly, and the read paths checked (`GetForTenantAsync`, `GetTemporalConfigAsync`, `GetFlowServerSettings`) scope lookups to the authenticated tenant context — no cross-tenant IDOR found. The new, unreported issue is the `request.TenantId` vs `validatedTenant.TenantId` mismatch above, which can silently defeat the isolation feature for affected tenant IDs. All other design gaps (isolation flag not gated on read, no revoke/rotate path, missing index, orphaned-record ordering, missing tests, SSRF on host input, log sanitization, non-atomic upsert) are already tracked in the 14 still-open threads from the prior review.

#### Test Coverage
Zero unit/integration tests exist for the three new production classes (`TenantTemporalConfig`, `TenantTemporalConfigRepository`, `TenantTemporalConfigService`) — this remains open from the prior review (thread on `TenantTemporalConfigService.cs:15`). This pass additionally found that the one touched test file only wires up the new mock dependency without configuring it, so the temporal-config code paths in `CreateTenant`/`UpdateTenant` are not actually exercised by the existing test suite (see Warnings above).

#### Performance
No N+1 queries, missed batching, or new blocking-I/O/sync-over-async patterns beyond what's already tracked (uncached `GetForTenantAsync` round-trip in `CertificateService`, the global `_connectionSemaphore` serializing lookups across tenants, no cache invalidation on config rotation). This pass surfaced one new issue: the search-attributes registration cache key collision described above. The previously-flagged missing early-return guard in `EnsureSearchAttributesRegisteredAsync` has been fixed by commit `fdb0d2d` — confirmed present and functioning at `TemporalClientService.cs:164-167`.

---

### Files Reviewed

| `XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs` | +114/-0 | 🔴 High | New service: encryption, validation, race condition |
| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs` | +67/-41 | 🔴 High | Client caching, search-attribute cache-key collision |
| `XiansAi.Server.Src/Shared/Services/TenantService.cs` | +41/-3 | 🔴 High | Tenant-ID sanitization mismatch (critical) |
| `XiansAi.Server.Src/Shared/Repositories/TenantTemporalConfigRepository.cs` | +51/-0 | 🟡 Medium | New repo: missing index, rethrow-on-error |
| `XiansAi.Server.Src/Shared/Services/CertificateService.cs` | +21/-2 | 🟡 Medium | Async flow-server settings retrieval |
| `XiansAi.Server.Src/Shared/Data/Models/TenantTemporalConfig.cs` | +44/-0 | 🟡 Medium | New model; cert/key fields |
| `XiansAi.Server.Src/Shared/Data/Models/Tenant.cs` | +6/-1 | 🟢 Low | New `UseSpecificTemporalNamespace` flag |
| `XiansAi.Server.Src/Features/AgentApi/Endpoints/SettingsEndpoints.cs` | +2/-2 | 🟢 Low | Endpoint now async |
| `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs` | +2/-0 | 🟢 Low | DI registration |
| `XiansAi.Server.Tests/UnitTests/Shared/Services/TenantServiceMetadataTests.cs` | +2/-0 | 🟢 Low | Mock injected but unconfigured |

XiansAi.Server.Src/Shared/Services/TenantService.cs
```diff
@@ -649,6 +658,32 @@ public async Task<ServiceResult<TenantCreatedResult>> CreateTenant(CreateTenantR
...
+                var hasCertificateKey = !string.IsNullOrWhiteSpace(request.TemporalCertificateKey);
+                if (hasCertificate != hasCertificateKey)
+                {
+                    return ServiceResult<TenantCreatedResult>.BadRequest(
+                        "Temporal certificate and certificate key must be provided together.");
+                }
+
+                _logger.LogInformation("Saving dedicated Temporal connection for tenant {TenantId}", LogSanitizer.Sanitize(request.TenantId));
+                await _tenantTemporalConfigService.SaveAsync(
+                    request.TenantId,
```
hasith, line 680: **Critical:** `_tenantTemporalConfigService.SaveAsync(...)` is called with the **unsanitized** `request.TenantId`, while the `Tenant` document itself is persisted under the **sanitized** `validatedTenant.TenantId` (produced by `SanitizeAndValidate()` a few lines earlier in this method). If `request.TenantId` contains characters stripped/normalized during sanitization, the two IDs diverge: the encrypted mTLS certificate/key config gets saved under an ID that later lookups (`GetByTenantIdAsync`, keyed off the canonical/sanitized `_tenantContext.TenantId`) will never match. The tenant then silently falls back to the shared/default Temporal namespace, defeating the isolation feature for that tenant, and an orphaned document containing an encrypted private key is left with no reachable owner.

Note: this is a distinct concern from the existing open thread at line 679 (that thread is about *ordering* — the config being saved before tenant creation succeeds; this finding is about the *tenant-ID value used* being unsanitized).

**Fix:** Use the already-in-scope `validatedTenant.TenantId` instead of the raw `request.TenantId`.


```suggestion
                    validatedTenant.TenantId,
```

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs
```diff
@@ -138,7 +162,9 @@ private TemporalConfig GetTemporalConfig(string tenantId)
     private async Task EnsureSearchAttributesRegisteredAsync(ITemporalClient client, string namespaceName)
     {
         if (_searchAttributesRegistered.TryGetValue(namespaceName, out var registered) && registered)
```
hasith, line 168: **Warning:** `_searchAttributesRegistered` (declared at line 21) is a process-wide `ConcurrentDictionary<string, bool>` keyed **only by namespace name**. Before this PR, namespace strings came from curated static configuration; this PR now lets any tenant admin supply an arbitrary Temporal `Namespace` via `TenantTemporalConfigService.SaveAsync` (called from `TenantService.CreateTenant`) pointing at a tenant-specific `Host`. If two tenants (or a tenant and the default cluster) pick the same namespace string against **different** Temporal hosts, the cache entry from the first tenant's registration check causes the second tenant's `EnsureSearchAttributesRegisteredAsync` call to short-circuit and skip verifying/registering search attributes against its own, different Temporal cluster — a silent cache-key collision that gets worse as tenant count grows and namespace names collide more often.

Note: this is distinct from the existing open thread at line 168 (that thread was about the early-return guard being missing entirely; the guard is now confirmed present/restored by commit fdb0d2d — this finding is about the cache *key* being too coarse, not about the guard's presence).

**Fix:** Key the cache by a composite of host+namespace (or tenantId+namespace) instead of namespace alone, e.g. `$"{config.FlowServerUrl}|{namespaceName}"`, so isolated per-tenant Temporal clusters can't share/mask each other's registration state.

XiansAi.Server.Tests/UnitTests/Shared/Services/TenantServiceMetadataTests.cs
```diff
@@ -27,6 +27,7 @@ public class TenantServiceMetadataTests
     private readonly Mock<ITenantCacheService> _cache = new();
     private readonly Mock<ITenantContext> _context = new();
     private readonly Mock<IRoleManagementService> _roles = new();
+    private readonly Mock<ITenantTemporalConfigService> _temporalConfigs = new();
```
hasith, line 30: **Warning:** The new `Mock<ITenantTemporalConfigService> _temporalConfigs` (line 30) is injected into `TenantService`'s constructor (line 62) but is never configured with any `.Setup(...)` return values anywhere in this test class. Any existing `CreateTenant`/`UpdateTenant` test that exercises a request with `UseSpecificTemporalNamespace = true` will silently invoke this unconfigured mock (returning `default`/`null`), which can mask behavioral bugs in the new temporal-config branches rather than actually exercising them — including the tenant-ID sanitization mismatch flagged separately in `TenantService.cs:680`.

**Fix:** Add explicit `.Setup(...)` calls for `_temporalConfigs` covering the temporal-namespace branches (e.g. verify it's called with the expected sanitized tenant ID), or add a dedicated test class for the new temporal-config creation/update paths.

XiansAi.Server.Src/Shared/Repositories/TenantTemporalConfigRepository.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Error retrieving Temporal config for tenant {TenantId}", LogSanitizer.Sanitize(tenantId));
+            throw;
```
hasith, line 34: **Suggestion:** `GetByTenantIdAsync` rethrows on any Mongo error, unlike the sibling `TenantOidcConfigRepository.GetByTenantIdAsync` (same shape), which returns `null` on error. This is not an information-disclosure risk (the global exception handler returns a generic 500 for unhandled exceptions), but it is a behavioral regression from the codebase's own convention: a transient Mongo hiccup now hard-fails every `GetFlowServerSettings()` call and every `TemporalClientService.GetClientAsync()` call for a tenant with dedicated Temporal config, instead of gracefully falling back to the default/static configuration on this availability-sensitive path.

Note: this is distinct from the existing open thread at line 29 (that thread is about the missing `tenant_id` index; this finding is about error-handling/fail-open behavior).

**Fix:** Consider whether fail-open (return `null`, fall back to static config) is more appropriate here, matching `TenantOidcConfigRepository`. Left as author judgment since it changes failure-mode semantics.

hasith (commented):
## PR Review Report

**PR:** Feature/temporal tenant isolation (#459)
**Author:** Diyanada99x
**Files Changed:** 10 | **+361** / **-49**
**Verdict:** REQUEST CHANGES

---

### Re-review delta

Reviewed 2 new commit(s) since the last review (`fdb0d2de`..`9f8474e`).
- ✅ Fixed: 0 previously-flagged issue(s) resolved
- ⏳ Still open: 19 carried-over issue(s)
- 🆕 New: 0 issue(s) introduced since the last review

---

### Existing review threads

- ✅ Appears addressed: 0 open thread(s) — will reply, leave open for original author
- ⏳ Still open: 19 open thread(s) — no reply (avoid spam)
- 🔇 Duplicates avoided: 6 finding(s) not re-posted (test-reviewer independently re-derived findings that are already tracked by open threads on `TenantService.cs:680`, `TenantTemporalConfigService.cs:15`, `TenantServiceMetadataTests.cs:30`, and the two `CertificateService.cs` log-sanitization lines — these were suppressed rather than duplicated)

---

### Summary

This revision (2 new commits since the last review) integrates `ITenantCacheService` into `CertificateService` and `TemporalClientService`. All four specialist reviewers (code, test, security, performance) re-scanned the full diff and confirmed no *new* issues were introduced — every substantive concern they found was already covered by one of the 19 still-open threads from the prior review pass. Notably, the security reviewer independently verified that `TenantTemporalConfigService` now encrypts the certificate/key before persisting and decrypts on read with `AuthenticationTagMismatchException` correctly rethrown — this looks like real progress on the previously-flagged plaintext-storage concern (thread on `TenantTemporalConfig.cs:31`), even though that thread's flagged property declarations are textually unchanged and so remain mechanically carried-over rather than auto-resolved. The still-open **CRITICAL** unsanitized-tenant-ID mismatch in `TenantService.cs:680` (`request.TenantId` vs `validatedTenant.TenantId`) has not been addressed and continues to block approval.

---

No *new* critical issues introduced since the last review. **1 previously-flagged CRITICAL issue remains open and unresolved:**

- [ ] `XiansAi.Server.Src/Shared/Services/TenantService.cs:680` — `_tenantTemporalConfigService.SaveAsync(...)` is still called with the unsanitized `request.TenantId` instead of `validatedTenant.TenantId`, so the encrypted Temporal config can be saved under an ID that later lookups (keyed off the sanitized tenant ID) never match — silently disabling namespace isolation for that tenant. (Open thread, unresolved — fix: use `validatedTenant.TenantId`.)

Also still open (mechanically carried-over, not re-verified as fixed by this pass): plaintext-at-rest concern on `XiansAi.Server.Src/Shared/Data/Models/TenantTemporalConfig.cs:31` — **worth a manual second look**, since the security reviewer's independent read of `TenantTemporalConfigService.cs` (encrypt-before-save at lines ~56-57, decrypt-on-read at ~103-104) suggests this may already be addressed at the service layer even though the flagged model property lines are unchanged.

---

### Warnings (Should Fix)

No new warnings introduced since the last review. 12 previously-flagged warnings/suggestions remain open across `TemporalClientService.cs`, `CertificateService.cs`, `TenantTemporalConfigService.cs`, `TenantTemporalConfigRepository.cs`, `TenantService.cs`, and `TenantServiceMetadataTests.cs` (see existing inline threads on this PR) — including the dropped early-return guard in `EnsureSearchAttributesRegisteredAsync` and the missing `createdBy`/host-validation checks in `SaveAsync`.

---

No new suggestions this pass — see existing open threads for previously-flagged items (missing `tenant_id` index, connection-semaphore contention, indefinite `ITemporalClient` caching, etc.).

---

### Review Details

#### Code Quality
No new code-quality issues found in this revision's changes; the `ITenantCacheService` integration into `CertificateService`/`TemporalClientService` is consistent with existing DI patterns.

#### Security
Verified the previously-flagged plaintext-storage issue now appears addressed at the service layer (`_encryption.Encrypt`/`Decrypt` around certificate/key, with `AuthenticationTagMismatchException` rethrown rather than swallowed). Tenant-cache keys are correctly scoped per tenant id; no cross-tenant leakage found in the new cache integration. No secrets found in new log statements. The unsanitized-tenant-ID CRITICAL (`TenantService.cs:680`) remains unresolved.

#### Test Coverage
No test changes accompany this revision's `ITenantCacheService` integration. The pre-existing gap remains: zero unit/integration tests for `TenantTemporalConfigService`, `TenantTemporalConfigRepository`, and the new `TenantService` temporal-config branches; the one touched test file only adds an unconfigured mock (already tracked in an open thread).

#### Performance
Confirmed the previously-flagged missing early-return guard in `EnsureSearchAttributesRegisteredAsync` is present in this revision (no regression), and that `ITenantCacheService`-backed lookups in `CertificateService`/`TemporalClientService` are `IMemoryCache`-backed and not re-executed redundantly. The `TenantTemporalConfig` Mongo lookup itself still lacks caching (already tracked) and the read-then-write `SaveAsync` race window is unchanged (already tracked).

---

### Files Reviewed

| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs` | +76/-36 | 🔴 High | `ITenantCacheService` integration, connection/client caching |
| `XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs` | +114/-0 | 🔴 High | New service; encrypts/decrypts mTLS cert & key |
| `XiansAi.Server.Src/Shared/Repositories/TenantTemporalConfigRepository.cs` | +51/-0 | 🟡 Medium | New Mongo repository for tenant temporal config |
| `XiansAi.Server.Src/Shared/Services/CertificateService.cs` | +27/-3 | 🟡 Medium | `ITenantCacheService` integration for flow server settings |
| `XiansAi.Server.Src/Shared/Services/TenantService.cs` | +41/-3 | 🔴 High | Tenant create/update now saves temporal config (unsanitized ID bug) |
| `XiansAi.Server.Src/Shared/Data/Models/TenantTemporalConfig.cs` | +44/-0 | 🟡 Medium | New model for cert/key + connection details |
| `XiansAi.Server.Src/Shared/Data/Models/Tenant.cs` | +6/-1 | 🟢 Low | Added `UseSpecificTemporalNamespace` property |
| `XiansAi.Server.Src/Shared/Configuration/SharedConfiguration.cs` | +2/-0 | 🟢 Low | DI registration |
| `XiansAi.Server.Src/Features/AgentApi/Endpoints/SettingsEndpoints.cs` | +3/-1 | 🟢 Low | Async flow server settings retrieval |
| `XiansAi.Server.Tests/UnitTests/Shared/Services/TenantServiceMetadataTests.cs` | +2/-0 | 🟢 Low | Adds unconfigured mock (see open thread) |

PR #463, opened 2026-08-10
Message roundtrip performance

hasith:


hasith (commented):
_This review was posted by a sub-agent outside the normal orchestration flow and has been retracted to avoid duplicate/conflicting threads. Its inline comments have been removed; its one unique finding (missing cache invalidation on delete in `MessageService`) has been folded into the canonical review below._

hasith (commented):
## PR Review Report

**PR:** Message roundtrip performance (#463)
**Author:** hasith
**Files Changed:** 16 | **+377** / **-52**
**Verdict:** REQUEST CHANGES

> **Note:** A sub-agent involved in producing this review posted a duplicate draft review directly to this PR outside the normal flow; it has been retracted (see the "COMMENTED" review above marked retracted) and its one unique finding folded into this report. Apologies for the noise.

---

### Existing review threads

- ⏳ Still open: 1 open thread(s) — no reply (avoid spam)
- 🔇 Duplicates avoided: 0 finding(s) not re-posted

---

### Summary
This PR delivers a well-motivated set of message-roundtrip performance optimizations — in-memory caching for thread IDs and incoming-origin/data, Temporal client reuse by endpoint, cluster-aware MongoDB connection settings, and a new supporting index — with generally clean, well-documented implementation. Two of the new caches have correctness/security gaps that should block merge: the Temporal client endpoint-reuse cache ignores per-tenant mTLS credentials (cross-tenant credential leakage), and the new incoming-origin cache is never invalidated on thread/message delete (stale data served after deletion). Several of the new caching paths also lack unit test coverage and have a couple of secondary correctness/design points (a cache-write-ordering race, a widened certificate-trust window, and sequential startup warm-up) worth addressing before or shortly after merge.

---

- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs:66` — Endpoint-keyed client cache (`BuildEndpointKey`, url+namespace only) ignores per-tenant mTLS certificate/key, so two tenants sharing a Temporal URL/namespace but with different certificates can end up sharing one tenant's authenticated connection.
  ```csharp
  // Current (problematic)
  private static string BuildEndpointKey(string flowServerUrl, string flowServerNamespace)
  {
      return $"{flowServerUrl}|{flowServerNamespace}";
  }

  // Fix
  private static string BuildEndpointKey(string flowServerUrl, string flowServerNamespace, string? certificateBase64)
  {
      var certFingerprint = certificateBase64 is null
          ? "no-cert"
          : Convert.ToHexString(SHA256.HashData(Convert.FromBase64String(certificateBase64)))[..16];
      return $"{flowServerUrl}|{flowServerNamespace}|{certFingerprint}";
  }
  ```
- [ ] `XiansAi.Server.Src/Shared/Services/MessageService.cs:519` — The new incoming-origin/data cache (`GetLastIncomingOriginAndDataCachedAsync`) is never invalidated when the underlying thread/messages are deleted. `DeleteThreadAsync` and `DeleteMessagesByTopicAsync` remove data with no knowledge of this cache, so a reply within the 10-minute cache window after a delete is auto-populated with stale `Origin`/`Data`.
  ```csharp
  // Fix
  private void InvalidateIncomingOriginCache(string threadId, string? scope = null)
  {
      var cacheKey = BuildIncomingOriginCacheKey(_tenantContext.TenantId, threadId, scope);
      _memoryCache.Remove(cacheKey);
  }
  // call from DeleteThreadAsync / DeleteMessagesByTopicAsync after the delete succeeds
  ```

---

### Warnings (Should Fix)

- [ ] `XiansAi.Server.Src/Features/AgentApi/Auth/ICertificateValidationCache.cs:127` — Cert-validation cache switched to sliding expiration with a 30-min absolute ceiling (was a fixed 10-min absolute expiration), widening the window a revoked certificate can remain trusted, especially across multiple instances that don't share revocation state.
- [ ] `XiansAi.Server.Src/Shared/Services/MessageService.cs:443` — `CacheIncomingOriginAndData` unconditionally overwrites the cache on every inbound call with no ordering/versioning; concurrent inbound messages to the same thread can leave a stale Origin/Data cached for up to 10 minutes, misrouting outbound replies.
- [ ] `XiansAi.Server.Src/Program.cs:439` — Per-tenant Temporal warm-up runs sequentially in a `foreach` and is awaited before the app serves traffic; startup time scales linearly with tenant count and one slow endpoint delays warming for all tenants after it.
- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs:18` — No unit tests for the new `_clientsByEndpoint` reuse logic; test coverage here would likely have caught the CRITICAL cross-tenant credential-sharing issue above.
- [ ] `XiansAi.Server.Src/Shared/Repositories/TenantRepository.cs:56` — `GetByTenantIdAsync`'s new nullable return contract has no explicit test asserting the null path when the tenant doesn't exist.

---

- [ ] `XiansAi.Server.Src/Features/AgentApi/Auth/ICertificateValidationCache.cs:46` — No unit tests for the new cache configuration-validation / expiration logic.
- [ ] `XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs:333` — No test for thread-ID cache invalidation on delete or concurrent create races.
- [ ] `XiansAi.Server.Src/Shared/Services/ActivationValidationService.cs:54` — No test validates that custom `Messaging:ActivationCacheMinutes` / `WorkflowTypeCacheMinutes` config values actually take effect.
- [ ] `XiansAi.Server.Src/Shared/Services/MessageService.cs:538` — `IncomingOriginCacheEntry` payloads are counted as size `1` regardless of actual size, sharing a size budget with unrelated caches (cert validation, thread IDs).
- [ ] `XiansAi.Server.Src/mongodb-indexes.yaml:122` — New `thread_incoming_origin_lookup` index substantially overlaps the existing `thread_scope_lookup` index; combined with the new in-memory cache reducing read pressure, the extra write-amplification cost may no longer be justified.
- [ ] `XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs:1392` — Trailing whitespace after closing brace.

---

### Review Details

#### Code Quality
Overall clean and well-documented — the timing helpers (`TimedAsync`), cluster-aware Mongo settings, and configurable cache durations are good patterns. Only a trailing-whitespace nit was found; no structural or naming issues.

#### Security
The core caching/tenant-scoping across `ConversationRepository`, `ActivationValidationService`, `MongoDbClientService`, and `TenantRepository` is sound, and demoting decrypted-message logging to `LogTrace` is a good improvement. The Temporal client endpoint-reuse cache is a real cross-tenant credential-sharing risk (CRITICAL, see above), and the certificate-validation cache's move to sliding expiration meaningfully widens the revoked-cert trust window (WARNING).

#### Test Coverage
Several new, concurrency-sensitive caching code paths (certificate validation cache, Temporal client endpoint reuse, MessageService origin/data cache, ConversationRepository thread-ID cache) ship with no new unit tests. Given the concrete bug found in the Temporal client reuse logic, test coverage for that path in particular should be prioritized. The `TenantRepository` nullable-return contract change also lacks a regression test for the null case.

#### Performance
The overall direction (reducing DB round-trips via in-memory caching, reusing Temporal clients, cluster-aware Mongo pooling) is sound and should meaningfully improve roundtrip latency. Two issues need attention: the incoming-origin cache has a staleness race under concurrent load (WARNING), and startup tenant warm-up is sequential rather than parallel (WARNING). The new Mongo index's marginal value versus its write-amplification cost, given the new cache layer, is worth re-examining (SUGGESTION).

---

### Files Reviewed

| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs` | +22/-1 | 🔴 High | Cross-tenant credential-sharing bug in new client-reuse cache |
| `XiansAi.Server.Src/Features/AgentApi/Auth/ICertificateValidationCache.cs` | +19/-3 | 🔴 High | Cert-validation cache expiration behavior change |
| `XiansAi.Server.Src/Shared/Services/MessageService.cs` | +130/-7 | 🟠 Medium | New in-memory origin/data cache, staleness race |
| `XiansAi.Server.Src/Program.cs` | +65/-4 | 🟠 Medium | Sequential tenant warm-up at startup |
| `XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs` | +58/-2 | 🟠 Medium | New thread-ID cache with invalidation |
| `XiansAi.Server.Src/Shared/Repositories/TenantRepository.cs` | +3/-1 | 🟡 Low-Medium | Nullable return contract change |
| `XiansAi.Server.Src/Shared/Data/MongoDbClientService.cs` | +40/-5 | 🟡 Low-Medium | Cluster-specific connection settings |
| `XiansAi.Server.Src/Shared/Services/ActivationValidationService.cs` | +19/-7 | 🟢 Low | Configurable cache durations |
| `XiansAi.Server.Src/mongodb-indexes.yaml` | +10/-0 | 🟢 Low | New compound index, overlaps existing index |
| `XiansAi.Server.Src/Shared/Data/MongoDBConfig.cs` | +5/-0 | 🟢 Low | New config options |
| `XiansAi.Server.Src/Shared/Exceptions/InvalidWorkflowIdentifierException.cs` | +6/-2 | 🟢 Low | Nullable identifier |
| `XiansAi.Server.Src/Features/AgentApi/Endpoints/MessagingEndpoints.cs` | +1/-1 | 🟢 Low | Minor change |
| `XiansAi.Server.Src/Features/UserApi/Endpoints/RestEndpoints.cs` | +3/-3 | 🟢 Low | Minor change |
| `XiansAi.Server.Src/Features/UserApi/Services/MongoChangeStreamService.cs` | +1/-1 | 🟢 Low | Minor change |
| `XiansAi.Server.Tests/UnitTests/Shared/Services/ActivationValidationServiceTests.cs` | +2/-2 | 🟢 Low | Test constructor update |
| `XiansAi.Server.Tests/XiansAi.Server.Tests.csproj` | +7/-0 | 🟢 Low | Framework reference addition |

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs
```diff
@@ -59,6 +59,18 @@ public async Task<ITemporalClient> GetClientAsync(string tenantId)
             }
 
             var config = GetTemporalConfig(tenantId);
+            var endpointKey = BuildEndpointKey(config.FlowServerUrl!, config.FlowServerNamespace!);
+
+            // Reuse an already-connected client for the same Temporal URL/namespace.
+            // Startup often warms "default" while requests use the real tenant id with the same root config.
+            if (_clientsByEndpoint.TryGetValue(endpointKey, out var sharedClient))
```
hasith, line 68: **Cross-tenant Temporal client/credential reuse.** The new endpoint cache key (`BuildEndpointKey`, line 121-124) is built only from `FlowServerUrl` + `FlowServerNamespace`, ignoring `config.CertificateBase64` / `config.PrivateKeyBase64`. Two tenants configured with the same Temporal URL/namespace but different (or no) mTLS client certificates will silently share the first tenant's authenticated connection — a cross-tenant credential/authorization leakage risk.

**Fix:** Include a fingerprint of the credential material in the cache key, e.g.:
```csharp
private static string BuildEndpointKey(string flowServerUrl, string flowServerNamespace, string? certificateBase64)
{
    var certFingerprint = certificateBase64 is null ? "no-cert" : Convert.ToHexString(SHA256.HashData(Convert.FromBase64String(certificateBase64)))[..16];
    return $"{flowServerUrl}|{flowServerNamespace}|{certFingerprint}";
}
```
and pass `config.CertificateBase64` at the call site (line 62), so clients are only shared when both endpoint and credential identity match.

*(pre-existing: no — introduced by this PR)*

XiansAi.Server.Src/Features/AgentApi/Auth/ICertificateValidationCache.cs
```diff
@@ -104,10 +118,14 @@ public void CacheValidation(string thumbprint, CachedCertificateValidation valid
...
         var cacheKey = GetCacheKey(thumbprint);
         
+        // Sliding expiration keeps actively used agent certificates warm (a busy agent never pays
+        // the revoke check + chain build + tenant lookup), while the absolute expiration caps how
+        // long a stale entry can survive.
         // Use Normal priority to allow proper cache eviction when under memory pressure
         // Set size to enable eviction policy when cache size limit is configured
         var cacheOptions = new MemoryCacheEntryOptions()
-            .SetAbsoluteExpiration(_cacheDuration)
+            .SetSlidingExpiration(_cacheDuration)
```
hasith, line 127: **Trust-window widening from sliding expiration.** The cert-validation cache switched from a fixed absolute expiration to `SetSlidingExpiration(_cacheDuration).SetAbsoluteExpiration(_cacheMaxDuration)` (lines 127-128). A continuously-used certificate can now stay trusted up to 3x longer (default 30 min ceiling vs. previous 10 min) without revalidation, widening the window in which a revoked/deleted certificate remains accepted — especially in multi-instance deployments where revocation doesn't broadcast to other instances' local cache.

**Fix:** If a bounded revalidation window matters for security, keep a lower absolute-only expiration, or explicitly document/lower the default `CertificateValidationCacheMaxDurationMinutes` and ensure revocation events invalidate the cache cluster-wide rather than relying on local TTL alone.

*(pre-existing: no — introduced by this PR)*

XiansAi.Server.Src/Shared/Services/MessageService.cs
```diff
@@ -397,14 +426,30 @@ public async Task<ServiceResult<string>> ProcessIncomingMessage(ChatOrDataReques
...
+        // so SignalWithStart does not depend on the Mongo insert completing first.
+        var saveTask = TimedAsync(
+            () => SaveMessage(threadId, request, MessageDirection.Incoming, messageType),
+            ms => saveMs = ms);
+        var signalTask = TimedAsync(
+            () => SignalWorkflowAsync(threadId, request, messageType),
+            ms => signalMs = ms);
+        await Task.WhenAll(saveTask, signalTask);
+
+        CacheIncomingOriginAndData(threadId, request);
```
hasith, line 439: **Cache staleness race under concurrent inbound messages.** `CacheIncomingOriginAndData` (called at line 443, defined at line 504) unconditionally overwrites the shared `IncomingOriginCacheEntry` cache entry on every inbound call. Under concurrent inbound messages to the same thread/scope, whichever request finishes last wins the cache write — not whichever message is actually newest by `created_at`. The DB path this cache replaces (`GetLastIncomingOriginAndDataAsync`) is always sorted by `created_at desc` and therefore authoritative; the cache has no such ordering, so a busy thread with rapid concurrent inbound traffic (exactly the scenario this PR optimizes for) can serve a stale Origin/Data pair for up to 10 minutes, misrouting outbound replies.

**Fix:** Store the persisted message's `CreatedAt` alongside `Origin`/`Data` and only overwrite the cache when the new value is newer (compare-and-set):
```csharp
private sealed record IncomingOriginCacheEntry(string? Origin, object? Data, DateTime CreatedAtUtc);

private void CacheIncomingOriginAndData(string threadId, ChatOrDataRequest request, DateTime createdAtUtc)
{
    var cacheKey = BuildIncomingOriginCacheKey(threadId, request.Scope);
    if (!_memoryCache.TryGetValue(cacheKey, out IncomingOriginCacheEntry? existing) || existing is null || createdAtUtc >= existing.CreatedAtUtc)
    {
        _memoryCache.Set(cacheKey, new IncomingOriginCacheEntry(request.Origin, request.Data, createdAtUtc), new MemoryCacheEntryOptions { Size = 1 }.SetAbsoluteExpiration(IncomingOriginCacheDuration));
    }
}
```
(Requires threading the persisted message's `CreatedAt` back from `SaveMessage` — not a single-line drop-in.)

*(pre-existing: no — introduced by this PR)*

XiansAi.Server.Src/Program.cs
```diff
@@ -424,22 +424,32 @@ private static void ValidateConfiguration(IConfiguration config)
...
-            // Get default tenant ID from configuration
-            var defaultTenantId = configuration["Tenants:default:TenantId"] ?? "default";
-            
-            // Attempt to connect - this validates the connection
-            await temporalClientService.GetClientAsync(defaultTenantId);
-            
-            _logger.LogInformation("Temporal connection validated successfully for tenant {TenantId}", defaultTenantId);
+            var tenantIdsToWarm = GetTemporalTenantIdsToWarm(configuration);
+
+            foreach (var tenantId in tenantIdsToWarm)
```
hasith, line 439: **Sequential tenant warm-up serializes startup.** `ValidateTemporalConnection` awaits each tenant's `GetClientAsync` one at a time inside a `foreach` (line 439), and this whole method is awaited before the app starts serving traffic. Startup/readiness time now scales linearly with the number of configured tenants, and a slow/unreachable Temporal endpoint for one tenant delays warming (and app start) for every tenant after it — even though each failure is individually caught.

**Fix:** Fire warm-up calls concurrently:
```csharp
var warmTasks = tenantIdsToWarm.Select(async tenantId =>
{
    try
    {
        await temporalClientService.GetClientAsync(tenantId);
        _logger.LogInformation("Temporal connection validated successfully for tenant {TenantId}", tenantId);
    }
    catch (Exception ex)
    {
        _logger.LogWarning(ex, "Failed to validate Temporal connection during startup for tenant {TenantId}. Workflows for this tenant may fail until Temporal is available.", tenantId);
    }
});
await Task.WhenAll(warmTasks);
```
Note `TemporalClientService.GetClientAsync` also serializes new-client creation behind a single global `SemaphoreSlim(1,1)` (pre-existing, unchanged by this PR) — for full benefit that would need to be scoped per-endpoint too.


```suggestion
            var warmTasks = tenantIdsToWarm.Select(async tenantId =>
            {
                try
                {
                    await temporalClientService.GetClientAsync(tenantId);
                    _logger.LogInformation("Temporal connection validated successfully for tenant {TenantId}", tenantId);
                }
                catch (Exception ex)
                {
                    _logger.LogWarning(ex,
                        "Failed to validate Temporal connection during startup for tenant {TenantId}. Workflows for this tenant may fail until Temporal is available.",
                        tenantId);
                }
            });
            await Task.WhenAll(warmTasks);
```

*(pre-existing: no — introduced by this PR)*

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs
```diff
@@ -16,6 +15,7 @@ public interface ITemporalClientService
 public class TemporalClientService : ITemporalClientService, IDisposable, IAsyncDisposable
 {
     private readonly ConcurrentDictionary<string, ITemporalClient> _clients = new();
+    private readonly ConcurrentDictionary<string, ITemporalClient> _clientsByEndpoint = new();
```
hasith, line 20: **No test coverage for the new endpoint-reuse cache.** The `_clientsByEndpoint` concurrent dictionary and endpoint-key reuse logic (lines 18, 62-72, 99, 111, 121-124, 313) have no unit tests — and, per the security finding above, this code has an actual correctness/security bug. New concurrency-sensitive shared-state logic like this needs test coverage for: client reuse across tenants with the same endpoint, cache-key collision behavior, cleanup on connection failure, and disposal clearing `_clientsByEndpoint`.

**Fix:** Add a `TemporalClientServiceTests` class mocking the Temporal client factory to cover these cases — a test asserting *distinct* clients for tenants with different certificates but the same URL/namespace would have caught the CRITICAL finding above.

*(pre-existing: no — introduced by this PR)*

XiansAi.Server.Src/Shared/Repositories/TenantRepository.cs
```diff
@@ -53,7 +53,7 @@ public async Task<Tenant> GetByIdAsync(string id, CancellationToken cancellation
         }, _logger, maxRetries: 3, baseDelayMs: 100, operationName: "GetTenantById");
     }
 
-    public async Task<Tenant> GetByTenantIdAsync(string tenantId, CancellationToken cancellationToken = default)
+    public async Task<Tenant?> GetByTenantIdAsync(string tenantId, CancellationToken cancellationToken = default)
```
hasith, line 56: **Untested null-return path.** `GetByTenantIdAsync` now returns `Tenant?` instead of throwing/returning a non-null default, but no test explicitly asserts the null path when the tenant doesn't exist. Since this is a public contract change (callers must now null-check), a regression test protects against a future revert.

**Fix:** Add a test case asserting `GetByTenantIdAsync` returns `null` for a non-existent `tenantId`, and verify call sites (e.g. anything using `WorkflowIdentifier`) handle the null correctly.

*(pre-existing: no — introduced by this PR)*

XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs
```diff
@@ -323,7 +330,19 @@ public ConversationRepository(
 
     public async Task<string> CreateOrGetThreadIdAsync(ConversationThread thread)
     {
-        return await MongoRetryHelper.ExecuteWithRetryAsync(async () =>
+        var cacheKey = BuildThreadIdCacheKey(thread.TenantId, thread.WorkflowId, thread.ParticipantId);
```
hasith, line 336: The new thread-ID cache in `CreateOrGetThreadIdAsync` (population at line 333, invalidation on delete at ~1355-1373) has no explicit test for cache invalidation on delete or for concurrent create-same-thread races. Consider an integration test covering: cache hit after first call, cache invalidated after `DeleteThreadAsync`, and concurrent calls for the same composite key not producing duplicate inserts.

*(pre-existing: no — introduced by this PR)*

XiansAi.Server.Src/Shared/Services/ActivationValidationService.cs
```diff
@@ -51,26 +51,38 @@ public class ActivationValidationService : IActivationValidationService
     private const string CacheKeyPrefix = "activation:validation:";
     // Per-agent list of registered workflow types (invalidatable by agent alone).
     private const string AgentWorkflowTypesCacheKeyPrefix = "activation:agent-workflow-types:";
-    private static readonly TimeSpan CacheDuration = TimeSpan.FromMinutes(2);
-    // Flow definitions are rarely changed at runtime — cache for 5 minutes to reduce DB load
-    // on every inbound message and adminapi /send, /listen, /history, /topics call.
-    private static readonly TimeSpan WorkflowTypeCacheDuration = TimeSpan.FromMinutes(5);
+    private const double DefaultCacheMinutes = 5;
```
hasith, line 54: Cache durations moved from hard-coded constants to configuration-driven values (`Messaging:ActivationCacheMinutes`, `Messaging:WorkflowTypeCacheMinutes`, lines 54-55, 65-66, 82-85), but the existing test only builds a default `IConfigurationBuilder` with no override, so it never exercises a custom duration. Consider a parameterized test that sets these keys via `AddInMemoryCollection` and asserts the configured value is actually used.

*(pre-existing: no — introduced by this PR)*

XiansAi.Server.Src/Shared/Services/MessageService.cs
```diff
@@ -424,7 +469,73 @@ private void ExtractWorkflowId(ChatOrDataRequest request)
...
+            cacheKey,
+            new IncomingOriginCacheEntry(origin, data),
+            new MemoryCacheEntryOptions()
+                .SetAbsoluteExpiration(IncomingOriginCacheDuration)
+                .SetSize(1));
+
+        return (origin, data);
+    }
 
+    private sealed record IncomingOriginCacheEntry(string? Origin, object? Data);
```
hasith: `IncomingOriginCacheEntry.Data` can hold an arbitrary caller-supplied payload but is counted as size `1` towards the shared `IMemoryCache` `SizeLimit` (same cache instance used by certificate validation and thread-ID caching). A workload with many distinct threads sending large `Data` payloads can grow real heap usage well beyond what the entry-count limit implies, and can crowd out unrelated entries sharing the same size budget.

**Fix:** Consider a separate size-limited `IMemoryCache` instance for this cache, or only cache `Origin` (the field that actually needs auto-population) and re-fetch `Data` from Mongo on the rare miss.

*(pre-existing: no — introduced by this PR)*

XiansAi.Server.Src/mongodb-indexes.yaml
```diff
@@ -117,6 +117,16 @@ conversation_message:
       scope: asc
       created_at: desc
     background: true
+  # perf: covers GetLastIncomingOriginAndDataAsync (tenant + thread + direction + scope + created_at)
+  # used when auto-populating Origin/Data on outbound replies and the in-memory cache misses.
+  - name: thread_incoming_origin_lookup
```
hasith, line 122: The new `thread_incoming_origin_lookup` compound index (`tenant_id, thread_id, direction, scope, created_at`) substantially overlaps the existing `thread_scope_lookup` index (`tenant_id, thread_id, scope, created_at`). This same PR also adds an in-memory cache in `MessageService` that removes most of the read pressure this index was meant to relieve, so the extra write-amplification cost (one more B-tree entry maintained per insert on a high-write-volume collection) may no longer be justified.

**Fix:** Verify with `explain()` whether `thread_scope_lookup` already serves `GetLastIncomingOriginAndDataAsync` acceptably now that reads are cache-served; drop the new index if so, or monitor write latency post-deploy if it's kept.

*(pre-existing: no — introduced by this PR)*

XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs
```diff
@@ -1335,7 +1389,7 @@ private void DecryptMessageText(ConversationMessage message)
                 // Not a valid Base64 string - this is plain text
                 _logger.LogDebug("Message {MessageId} is not encrypted (invalid Base64), treating as plain text", LogSanitizer.Sanitize(message.Id));
                 // Leave message.Text as-is
-            }
+            }
```
hasith, line 1402: Trailing whitespace after the closing brace (`}   `) — minor cleanup, no functional impact.


```suggestion
            }
```

XiansAi.Server.Src/Features/AgentApi/Auth/ICertificateValidationCache.cs
```diff
@@ -48,6 +48,7 @@ public class MemoryCertificateValidationCache : ICertificateValidationCache
     private readonly IMemoryCache _cache;
     private readonly ILogger<MemoryCertificateValidationCache> _logger;
     private readonly TimeSpan _cacheDuration;
+    private readonly TimeSpan _cacheMaxDuration;
```
hasith, line 51: No unit tests were added for the new `MemoryCertificateValidationCache` configuration-validation and sliding/absolute expiration logic (lines 51, 69-77, 127-128). Given the trust-window behavior change flagged above, a test asserting the configured ceiling is honored (and that `MaxDuration < Duration` is validated/clamped at construction) would help lock in intended behavior.

*(pre-existing: no — introduced by this PR)*

XiansAi.Server.Src/Shared/Services/MessageService.cs
```diff
@@ -424,7 +469,73 @@ private void ExtractWorkflowId(ChatOrDataRequest request)
...
+            cacheKey,
+            entry,
+            new MemoryCacheEntryOptions()
+                .SetAbsoluteExpiration(IncomingOriginCacheDuration)
+                .SetSize(1));
+    }
+
+    private async Task<(string? Origin, object? Data)> GetLastIncomingOriginAndDataCachedAsync(string threadId, string? scope)
+    {
+        var cacheKey = BuildIncomingOriginCacheKey(_tenantContext.TenantId, threadId, scope);
```
hasith: **CRITICAL:** The new "last incoming origin/data" cache (`GetLastIncomingOriginAndDataCachedAsync`, keyed by `tenantId:threadId:scope`, 10-minute absolute expiration) is never invalidated when the underlying data changes. `DeleteThreadAsync` and `DeleteMessagesByTopicAsync` delete messages/threads with no knowledge of this cache. A reply issued for that thread/scope within the 10-minute window after a delete will be auto-populated with stale `Origin`/`Data` served straight from cache, silently routing it based on deleted/stale state — a functional regression introduced by this performance optimization.

**Fix:** Invalidate the cache entry from the delete paths:
```csharp
private void InvalidateIncomingOriginCache(string threadId, string? scope = null)
{
    var cacheKey = BuildIncomingOriginCacheKey(_tenantContext.TenantId, threadId, scope);
    _memoryCache.Remove(cacheKey);
}
```
Call this from `DeleteThreadAsync` and `DeleteMessagesByTopicAsync` after the underlying delete succeeds.

*(pre-existing: no — introduced by this PR)*

hasith (commented):
## PR Review Report

**PR:** Message roundtrip performance (#463)
**Author:** hasith
**Files Changed:** 19 | **+680** / **-55**
**Verdict:** NEEDS DISCUSSION

---

### Re-review delta
Reviewed 1 new commit since the last review (`2629e53`..`d234807`).
- ✅ Fixed: 2 previously-flagged issue(s) resolved (per automated fid tracking)
- ⏳ Still open: 11 carried-over issue(s)
- 🆕 New: 3 issue(s) introduced since the last review

> **Note on the "cross-tenant Temporal client/credential reuse" thread** (`TemporalClientService.cs:68`, originally CRITICAL): this run's automated fid-based re-verification (Gate B) keeps it in the *carried-over* bucket because the specific line it anchors to (`if (_clientsByEndpoint.TryGetValue(...))`) is unchanged text-for-text. However, manual inspection of the current code shows the underlying vulnerability **is fixed**: `BuildEndpointKey` (lines 128–145) now folds a SHA-256 fingerprint of `CertificateBase64`/`PrivateKeyBase64` into the cache key via a new `BuildCredentialFingerprint` helper, so tenants with different mTLS credentials no longer share a connection. This is a known limitation of anchor-text-based fid matching (the fix landed in a different line than the flagged one) — left open per the plugin's conservative policy rather than auto-resolved, but a maintainer can safely resolve that thread. The other original CRITICAL (`MessageService.cs` — incoming-origin cache never invalidated on delete) is correctly tracked as **Fixed**: `ConversationRepository.cs` now calls `InvalidateThread`/`InvalidateScope` on all three delete paths (`DeleteThreadAsync`, `DeleteMessagesByThreadId`, `DeleteMessagesByWorkflowParticipantAndScope`).

---

### Existing review threads
- ⏳ Still open: 1 open thread(s) — no reply (avoid spam)
- 🔇 Duplicates avoided: 2 finding(s) not re-posted (already tracked as carried-over plugin threads: `TenantRepository.cs:56`, `ConversationRepository.cs:1402`)

---

### Summary
This re-review covers the latest commit (`refactor(tenant-repository): update GetByTenantIdAsync ... return nullable Tenant`) on top of the prior round. Both previously-flagged **CRITICAL** issues appear resolved in the current code: the Temporal endpoint-reuse cache now fingerprints mTLS credentials into its cache key, and the incoming-origin cache is now invalidated on all thread/message delete paths. No new critical or security issues were found by the code-quality, security, or performance specialists on this pass. The test-quality reviewer flagged a few remaining coverage gaps — some genuinely new (concurrency/edge-case tests for the new `IncomingOriginCache`, and a test for the new configurable cache-duration path in `ActivationValidationService`), and two that duplicate already-open threads from the prior round. Several WARNING/SUGGESTION-level items from the prior review (widened certificate trust window, a cache-write-ordering race, sequential Temporal warm-up at startup, missing tests for the endpoint-reuse cache) remain unaddressed and are unaffected by this commit.

---

No critical issues found. (Both prior CRITICAL findings — cross-tenant Temporal credential reuse, and missing cache invalidation on thread/message delete — appear fixed in the current code; see the re-review delta note above.)

---

### Warnings (Should Fix)

*(Carried over from the prior review — already tracked as open plugin threads, not re-posted)*

- [ ] `XiansAi.Server.Src/Features/AgentApi/Auth/ICertificateValidationCache.cs:127` — Cert-validation cache switched to sliding expiration with a 30-min absolute ceiling, widening the window a revoked certificate can remain trusted.
- [ ] `XiansAi.Server.Src/Shared/Services/MessageService.cs:439` — `CacheIncomingOriginAndData` unconditionally overwrites the cache on every inbound call with no ordering/versioning; concurrent inbound messages can leave a stale Origin/Data cached.
- [ ] `XiansAi.Server.Src/Program.cs:439` — Per-tenant Temporal warm-up runs sequentially and is awaited before the app serves traffic.
- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs:20` — No unit tests for the new endpoint-reuse cache (`_clientsByEndpoint`).
- [ ] `XiansAi.Server.Src/Shared/Repositories/TenantRepository.cs:56` — No test asserting the null path for the new nullable `GetByTenantIdAsync` contract.

**New this pass:**

- [ ] `XiansAi.Server.Tests/UnitTests/Shared/Services/IncomingOriginCacheTests.cs:137` — No test covers null/empty `tenantId`/`threadId` inputs.
- [ ] `XiansAi.Server.Tests/UnitTests/Shared/Services/IncomingOriginCacheTests.cs:137` — No test exercises concurrent `Set`/`Get`/`InvalidateThread` calls against `_threadEvictionSources`.

---

*(Carried over — already tracked as open plugin threads, not re-posted)*

- [ ] `XiansAi.Server.Src/Features/AgentApi/Auth/ICertificateValidationCache.cs:51` — No unit tests for the new configuration-validation/expiration logic.
- [ ] `XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs:336` — No test for thread-ID cache invalidation on delete or concurrent create races.
- [ ] `XiansAi.Server.Src/Shared/Services/ActivationValidationService.cs:54` — No test validates custom `Messaging:ActivationCacheMinutes`/`WorkflowTypeCacheMinutes` config values take effect.
- [ ] `XiansAi.Server.Src/mongodb-indexes.yaml:122` — New `thread_incoming_origin_lookup` index may overlap with existing indexes now that reads are cache-served.
- [ ] `XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs:1402` — Trailing whitespace after a closing brace.

**New this pass:**

- [ ] `XiansAi.Server.Tests/UnitTests/Shared/Services/ActivationValidationServiceTests.cs:32` — No test verifies configured `Messaging:ActivationCacheMinutes`/`WorkflowTypeCacheMinutes` values are actually honored (constructor now accepts `IConfiguration` but every test passes an empty config).

---

### Review Details

#### Code Quality
No new code-quality issues found by the specialist reviewer on this pass; nothing outstanding beyond what's already tracked in open threads.

#### Security
No new vulnerabilities found. Both previously-flagged CRITICAL findings (cross-tenant Temporal credential/connection reuse; stale incoming-origin cache surviving deletes) were independently re-verified against the current code and confirmed fixed — credential fingerprinting is now folded into `BuildEndpointKey`, and `ConversationRepository` now calls `InvalidateThread`/`InvalidateScope` on every delete path. Cache key tenant-isolation for the new `IncomingOriginCache` and the `ConversationRepository` thread-id cache was also verified.

#### Test Coverage
The new `IncomingOriginCache` has decent happy-path coverage (137-line test file) but lacks null/empty-input and concurrency tests for its `ConcurrentDictionary`-backed eviction-source recycling. `ActivationValidationServiceTests` doesn't exercise the new configurable cache-duration path. Two pre-existing coverage gaps from the prior round (nullable `GetByTenantIdAsync`, endpoint-reuse cache) remain unaddressed.

#### Performance
No new performance issues found. Cache-key namespacing across `IncomingOriginCache`, the `ConversationRepository` thread-id cache, `ActivationValidationService`, and others was verified to be collision-free on the shared `IMemoryCache`. The per-thread `CancellationTokenSource` dictionary in `IncomingOriginCache` is self-bounding (removed on invalidation or self-cancel). Previously-flagged items (cache-write-ordering race, sequential Temporal warm-up, redundant Mongo index) are unchanged by this commit.

---

### Files Reviewed

| `XiansAi.Server.Src/Shared/Services/IncomingOriginCache.cs` | +150 | 🟡 Medium | New caching service; core of this PR |
| `XiansAi.Server.Src/Shared/Services/MessageService.cs` | +100/-18 | 🟡 Medium | Cache integration for message processing |
| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs` | +43/-4 | 🟡 Medium | Client reuse across tenants; credential fingerprint fix verified |
| `XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs` | +65/-5 | 🟡 Medium | Thread-id caching, delete-path cache invalidation |
| `XiansAi.Server.Src/Program.cs` | +64/-5 | 🟡 Medium | Startup Temporal warm-up |
| `XiansAi.Server.Src/Features/AgentApi/Auth/ICertificateValidationCache.cs` | +19/-3 | 🟡 Medium | Sliding-expiration cache change |
| `XiansAi.Server.Src/Shared/Data/MongoDbClientService.cs` | +42/-3 | 🟢 Low | Conditional cluster settings |
| `XiansAi.Server.Src/Shared/Services/ActivationValidationService.cs` | +23/-3 | 🟢 Low | Configurable cache durations |
| `XiansAi.Server.Src/Shared/Repositories/TenantRepository.cs` | +2/-2 | 🟢 Low | Nullable return contract |
| `XiansAi.Server.Src/mongodb-indexes.yaml` | +10 | 🟢 Low | New compound index |
| `XiansAi.Server.Tests/UnitTests/Shared/Services/IncomingOriginCacheTests.cs` | +137 | 🟢 Low | New tests |
| …and 8 more files | +25/-12 | 🟢 Low | Config, minor endpoint/logging changes, test project setup |

---

XiansAi.Server.Src/Shared/Services/MessageService.cs
hasith: ✅ Resolved as of `d234807d8b8e5d801f1b6bff185c3bae7bfde0c4`. This finding no longer reproduces against the current head.

XiansAi.Server.Src/Shared/Services/MessageService.cs
hasith: ✅ Resolved as of `d234807d8b8e5d801f1b6bff185c3bae7bfde0c4`. This finding no longer reproduces against the current head.

XiansAi.Server.Tests/UnitTests/Shared/Services/IncomingOriginCacheTests.cs
```diff
@@ -0,0 +1,137 @@
...
+    public void InvalidateScope_DropsDefaultScope_WhenScopeIsNull()
+    {
+        var cache = BuildCache();
+        cache.Set(TenantId, ThreadId, null, new IncomingOriginData("app:web", null));
+
+        cache.InvalidateScope(TenantId, ThreadId, null);
+
+        Assert.Null(cache.Get(TenantId, ThreadId, null));
+    }
+}
```
hasith, line 137: **Issue:** `IncomingOriginCacheTests` doesn't cover null/empty `tenantId`/`threadId` inputs to `Get`, `Set`, `InvalidateThread`, and `InvalidateScope`. The production code appears to handle these gracefully, but that behavior is currently unverified by any test, so a future change could silently break it.

**Fix:** Add cases such as:
```csharp
[Fact]
public void Get_ReturnsNull_WhenTenantIdIsEmpty()
{
    var cache = BuildCache();
    Assert.Null(cache.Get("", ThreadId, null));
}
```

*(pre-existing: no — new test file introduced by this PR)*

hasith, line 137: **Issue:** `_threadEvictionSources` is a `ConcurrentDictionary` and `GetOrCreateEvictionSource` contains retry logic for concurrent cancellation, but no test exercises concurrent `Set`/`Get`/`InvalidateThread` calls racing against each other. This is exactly the kind of logic that looks correct on a single thread and breaks under real concurrency (the scenario this cache exists to handle, given it's used on the message-processing hot path).

**Fix:** Add a test that spins up several concurrent tasks calling `Set` and `InvalidateThread` on the same thread id and asserts no exceptions are thrown and the cache ends in a consistent state (e.g. via `Task.WhenAll(Enumerable.Range(0, N).Select(_ => Task.Run(...)))`).

*(pre-existing: no — new test file introduced by this PR)*

XiansAi.Server.Tests/UnitTests/Shared/Services/ActivationValidationServiceTests.cs
```diff
@@ -27,7 +28,8 @@ public ActivationValidationServiceTests()
             _activationRepository.Object,
             _flowDefinitionRepository.Object,
             _cache,
-            NullLogger<ActivationValidationService>.Instance);
+            NullLogger<ActivationValidationService>.Instance,
+            new ConfigurationBuilder().Build());
```
hasith, line 32: **Issue:** The constructor now takes an `IConfiguration` for cache-duration overrides (`Messaging:ActivationCacheMinutes`, `Messaging:WorkflowTypeCacheMinutes`), but every test in this file builds an empty `ConfigurationBuilder().Build()`, so the custom-duration code path is never exercised.

**Fix:** Add a test that constructs the service with `new ConfigurationBuilder().AddInMemoryCollection(new Dictionary<string,string?> { ["Messaging:ActivationCacheMinutes"] = "5" }).Build()` and asserts the configured duration is used (e.g. by observing cache-expiry behavior or exposing it for verification).

*(pre-existing: no — introduced by this PR)*

PR #470, opened 2026-08-11
44 when an agent is deleted we need to give user the option to delete all messages data etc

### LINKED ISSUE #44: When an agent is deleted, we need to give user the option to delete all messages, data etc.
In admin API, when activation is deleted, we should be pass flags letting user delete all associated entities like knowledge, data, logs, secrets, connections, etc.

in AS ui should have a check boxes

Diyanada99x:


hasith:
🔍 PR Review in Progress

Claude Code is analyzing this pull request. The review will be posted here shortly.

PR Reviewer (1.9.9)

hasith (commented):
## PR Review Report

**PR:** #470 — 44 when an agent is deleted we need to give user the option to delete all messages data etc
**Author:** Diyanada99x
**Files Changed:** 30 | **+1045** / **-29**
**Verdict:** REQUEST CHANGES

---

### Existing review threads
- ⏳ Still open: 10 open thread(s) — no reply (avoid spam)
- 🔇 Duplicates avoided: 0 finding(s) not re-posted (reviewers were instructed to skip the already-flagged "generic catch clause" instances on `ActivationService.cs`, `AdminDataService.cs`, and `AdminKnowledgeEndpoints.cs`)

---

### Summary
This PR adds a large, high-risk admin surface for permanently deleting tenant/agent/activation data — 9 new Admin API endpoints and cascading-delete logic spanning ~15 services and repositories, plus (in the latest commit) force-deletion of all activations when a tenant is removed. Tenant isolation on the new endpoints is solid (pre-existing `TenantRouteScopeFilter` enforces it) and Mongo filters are injection-safe, but the change ships with essentially zero test coverage for irreversible destructive operations, and the force-delete flow in `AgentService` can report failure after some activations — and their data — have already been permanently deleted.

---

- [ ] `XiansAi.Server.Src/Shared/Services/AgentService.cs:172` — Force-delete can report an overall failure ("Agent was not deleted — please retry") *after* already permanently deleting a subset of the agent's activations and all their cascaded data. A caller has no reliable way to tell what was actually destroyed before retrying.
  ```csharp
  // Current
  return ServiceResult<AgentDeletionResult>.Conflict(
      $"Deleted {deletedActivations} of {activations.Count} activation(s) for agent '{agentName}'. " +
      $"Failed: {string.Join("; ", failedActivations)}. Agent was not deleted — please retry.");

  // Fix: return which activations were actually destroyed, and make retry
  // idempotent (skip already-deleted activations) instead of re-attempting them.
  ```

- [ ] **No test coverage for the entire new destructive-deletion feature.** ~1,045 lines of new cascading/bulk-deletion logic (9 new Admin API endpoints + cascade logic in `ActivationService.CleanupActivationDataAsync`, `AgentService` force-delete, `TenantService.ForceDeleteActivationsForTenantAsync`, `TemplateService.ForceDeleteActivationsAcrossTenantsAsync`) ship with only an unrelated 8-line test diff. Given every one of these paths performs irreversible deletion, add integration tests for happy-path + count verification, empty-activation, and partial-failure behavior before merging. See inline comment on `ActivationService.cs:764` for the full list of untested paths.

---

### Warnings (Should Fix)

- [ ] `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminKnowledgeEndpoints.cs:1240` — New activation-wide knowledge-delete endpoint returns the raw `ex.Message` to the caller on failure (information disclosure), unlike the sibling endpoints added in this same PR which delegate to `ServiceResult<T>`/`.ToHttpResult()` and never leak exception text.
- [ ] `XiansAi.Server.Src/Shared/Services/ActivationService.cs:764` — `CleanupActivationDataAsync` runs 7 independent cleanup calls as fully sequential `await`s; parallelize with `Task.WhenAll` since none depend on each other, especially since this method is invoked in tight loops by the force-delete flows below.
- [ ] `XiansAi.Server.Src/Shared/Services/AgentService.cs:144` — Force-delete processes activations strictly sequentially (~9 round trips per activation); the identical pattern is duplicated in `TenantService.cs:1017` and `TemplateService.cs:297`. Use bounded concurrency and consider extracting the shared loop.
- [ ] `XiansAi.Server.Src/Shared/Services/MessageService.cs:705` — `DeleteMessagesAsync` deletes matching threads one at a time in a loop instead of a bulk `DeleteManyAsync`, unlike every other repository added in this PR.
- [ ] `XiansAi.Server.Src/Shared/Repositories/FeedbackRepository.cs:208` — New delete filter queries by `agent_name` + `workflow_id`, but `message_feedback` has no index covering either field, forcing a full tenant-collection scan.

---

- [ ] `XiansAi.Server.Src/Shared/Services/ActivationService.cs:730` — Best-effort cleanup failures are only logged as warnings with no reconciliation path; consider surfacing partial-cleanup failures in the response or tracking them for follow-up.
- [ ] `XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs:448` — `GetByTenantAndActivationAsync` appears to be unused dead code and, if ever called, would force a full tenant-wide scan (no agent-scoping).
- [ ] `XiansAi.Server.Src/Shared/Services/AppIntegrationService.cs:770` — Per-webhook deletes run sequentially; consider bounded-concurrency fan-out for agents/activations with many builtin webhooks.
- [ ] `XiansAi.Server.Src/Shared/Services/TenantService.cs:995` — Tenant/template-wide force-delete cascades have no dry-run/preview mode and don't log the invoking admin's identity; worth adding given the blast radius.

---

### Review Details

#### Code Quality
The new admin-deletion surface is large but broadly consistent: most new service methods follow the existing `ServiceResult<T>` + try/catch + `LogSanitizer` conventions, and the three new bulk-webhook-delete methods correctly share one private helper rather than duplicating logic. The main structural weakness is the "deactivate + delete + cascade" activation loop being copy-pasted across `AgentService`, `TenantService`, and `TemplateService` (see Warnings) instead of being extracted once. One new endpoint (`AdminKnowledgeEndpoints`) diverges from the pattern the other new endpoints establish by leaking exception details.

#### Security
Tenant isolation is enforced correctly on every new endpoint via the pre-existing `TenantRouteScopeFilter`/`AdminTenantScopeGuard`, and Mongo filter construction is injection-safe (strongly-typed `Builders<T>.Filter`, `Regex.Escape()` applied before building regex filters) — no cross-tenant IDOR or NoSQL-injection issues found. The residual risk is in partial-failure semantics of the new "force delete" cascades: cleanup failures are swallowed to warning-level logs with no reconciliation, and (more seriously) the `AgentService` force-delete path can report failure after some data is already irreversibly gone (Critical, above). The `AdminKnowledgeEndpoints` exception-message leak is a real but narrow information-disclosure issue on an authenticated admin-only endpoint.

#### Test Coverage
This is the biggest gap in the PR: a large, irreversible, cross-cutting deletion feature ships with no meaningful new test coverage. Every new endpoint and every new cascade/orchestration method is currently unverified by automated tests. See Critical Issues above.

#### Performance
Several new deletion paths use sequential loops where bulk operations or concurrent fan-out would be safe and considerably faster: the 7-call cleanup cascade, the three duplicated "deactivate + delete each activation" loops, the message-thread delete loop, and the webhook-delete loop. None of these are correctness bugs, but they compound into real timeout risk for tenants/agents/templates with many activations. A newly-added feedback delete filter also lacks a supporting index.

---

### Files Reviewed

| `XiansAi.Server.Src/Shared/Services/AgentService.cs` | +75/-6 | 🔴 High | Force-delete cascade, misleading partial-failure response |
| `XiansAi.Server.Src/Shared/Services/ActivationService.cs` | +114/-0 | 🔴 High | Core cascading-cleanup method, sequential awaits, no tests |
| `XiansAi.Server.Src/Shared/Services/TenantService.cs` | +103/-0 | 🔴 High | Tenant-wide force-delete cascade |
| `XiansAi.Server.Src/Shared/Services/TemplateService.cs` | +115/-0 | 🟡 Medium | Template-wide force-delete cascade, duplicated loop pattern |
| `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminKnowledgeEndpoints.cs` | +25/-0 | 🟡 Medium | New endpoint leaks exception message |
| `XiansAi.Server.Src/Shared/Services/AppIntegrationService.cs` | +79/-0 | 🟡 Medium | Bulk webhook delete, sequential fan-out |
| `XiansAi.Server.Src/Shared/Services/MessageService.cs` | +47/-1 | 🟡 Medium | N+1 delete loop instead of bulk delete |
| `XiansAi.Server.Src/Shared/Repositories/FeedbackRepository.cs` | +23/-0 | 🟡 Medium | New delete filter missing supporting index |
| `XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs` | +53/-0 | 🟢 Low | New method appears unused |
| `XiansAi.Server.Src/Shared/Services/AdminDataService.cs` | +34/-0 | 🟢 Low | New delete-by-activation method |
| `XiansAi.Server.Src/Shared/Services/AdminLogsService.cs` | +27/-0 | 🟢 Low | New delete-by-activation method |
| `XiansAi.Server.Src/Shared/Services/AdminMetricsService.cs` | +32/-0 | 🟢 Low | New delete-by-activation method |
| `XiansAi.Server.Src/Shared/Services/FeedbackService.cs` | +26/-0 | 🟢 Low | New delete-by-activation method |
| `XiansAi.Server.Src/Shared/Services/KnowledgeService.cs` | +16/-0 | 🟢 Low | New delete-by-activation method |
| `XiansAi.Server.Src/Shared/Repositories/KnowledgeRepository.cs` | +40/-0 | 🟢 Low | New bulk-delete repository methods |
| `XiansAi.Server.Src/Shared/Repositories/UsageEventRepository.cs` | +19/-1 | 🟢 Low | New bulk-delete repository method |
| `XiansAi.Server.Src/Shared/Repositories/AppIntegrationRepository.cs` | +20/-0 | 🟢 Low | New lookup methods for webhook cleanup |
| `XiansAi.Server.Src/Features/WebApi/Repositories/LogRepository.cs` | +19/-0 | 🟢 Low | New bulk-delete repository method |
| `...and 12 more files` (Admin endpoint wiring, tests) | +\~130/-22 | 🟢 Low | Thin endpoint wiring + `TenantServiceMetadataTests.cs` |

XiansAi.Server.Src/Shared/Services/AgentService.cs
```diff
@@ -119,20 +123,73 @@ public async Task<ServiceResult<AgentDeletionResult>> DeleteAgentAsync(string ag
...
+                            failedActivations.Add($"Failed to delete '{activation.Name}': {deleteResult.ErrorMessage}");
+                        }
+                    }
+
+                    if (failedActivations.Count > 0)
+                    {
+                        _logger.LogWarning(
+                            "Aborting delete of agent {AgentName} - {FailedCount} of {TotalCount} activation(s) failed to clean up",
+                            LogSanitizer.Sanitize(agentName), failedActivations.Count, activations.Count);
+                        return ServiceResult<AgentDeletionResult>.Conflict(
```
hasith: **Critical:** Force-delete can report failure *after* already permanently deleting some activations (and all their cascaded data — messages, documents, knowledge, logs, metrics, feedback).

In `DeleteAgentAsync`'s `forceDelete` path, the per-activation loop (`XiansAi.Server.Src/Shared/Services/AgentService.cs:144`) calls `_activationService.DeleteActivationAsync(activation.Id)` for each activation — which is **irreversible** (it triggers `CleanupActivationDataAsync`). If a *later* activation in the loop fails, the method returns:
```csharp
return ServiceResult<AgentDeletionResult>.Conflict(
    $"Deleted {deletedActivations} of {activations.Count} activation(s) for agent '{agentName}'. " +
    $"Failed: {string.Join("; ", failedActivations)}. Agent was not deleted — please retry.");
```
The message does state a count, which is good, but the overall operation is reported as a `Conflict`/failure while data for the successfully-processed activations is already gone. A caller retrying the same `forceDelete=true` call has no way to know which activations were already destroyed vs. which still need cleanup, and the "Agent was not deleted — please retry" framing reads as "nothing happened" rather than "partial, irreversible data loss occurred."

**Fix:** Either (a) return the list of successfully-deleted vs. failed activation names/IDs in the response body so the caller can reconcile state, or (b) make the retry path skip already-deleted activations so a second call safely completes the remainder instead of re-attempting them.

XiansAi.Server.Src/Shared/Services/ActivationService.cs
```diff
@@ -729,4 +754,93 @@ await _webhookEventPublisher.PublishAsync(
...
         }
     }
+
+    /// <summary>
+    /// Best-effort cleanup of all data scoped to an activation: conversation threads/messages,
+    /// documents, knowledge, logs, schedules, app integrations, usage metrics, and feedback.
+    /// Each resource type is cleaned up independently — a failure in one does not block the others
+    /// or the activation record deletion that follows.
+    /// </summary>
+    private async Task CleanupActivationDataAsync(AgentActivation activation)
```
hasith: **Critical: no test coverage for the entire new destructive-deletion feature.** This PR adds ~1,045 lines of new cascading/bulk-deletion logic across 9 new Admin API endpoints and ~15 services/repositories, all of which perform **irreversible** data deletion, but only one 8-line diff touches a test file (`TenantServiceMetadataTests.cs`, unrelated to deletion).

Untested new destructive code paths include:
- `CleanupActivationDataAsync` (this method) — the core best-effort cascade run on every activation delete
- `AgentDeletionService.DeleteAgentAsync` force-delete flow (`AgentService.cs:144`)
- `TenantService.ForceDeleteActivationsForTenantAsync` (`TenantService.cs:995`)
- `TemplateService.ForceDeleteActivationsAcrossTenantsAsync` (`TemplateService.cs:275`)
- All 7 new `DeleteXByActivationAsync` admin endpoints (messages, documents, feedback, knowledge, logs, metrics, webhooks)

**Fix:** Add integration tests (the repo already has `XiansAi.Server.Tests/IntegrationTests/AdminApi`) covering at minimum: happy-path deletion + count verification, empty-activation (nothing to delete), partial-failure behavior (one resource type fails — others should still be cleaned up and the activation/agent should still delete), and — for the force-delete flows — verification that a failure partway through is reported accurately (see the related finding on `AgentService.cs:172`).

XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminKnowledgeEndpoints.cs
```diff
@@ -1220,6 +1220,31 @@ Same as the original knowledge item. User must have permission to modify the age
...
+        {
+            try
+            {
+                var deletedCount = await knowledgeService.DeleteAllByAgentAndActivationForTenantAsync(tenantId, agentName, activationId);
+                return Results.Ok(new { message = $"Deleted {deletedCount} knowledge item(s)", deletedCount });
+            }
+            catch (Exception ex)
+            {
+                return Results.Json(
+                    new { error = "Internal server error", message = ex.Message },
```
hasith: **Information disclosure:** the new activation-wide knowledge-delete endpoint returns the raw exception message to the caller on failure:
```csharp
catch (Exception ex)
{
    return Results.Json(
        new { error = "Internal server error", message = ex.Message },
        statusCode: 500);
}
```
This can leak internal details (Mongo error text, field/collection names, occasionally connection info) to any caller who can trigger an exception on this admin endpoint (e.g. a malformed `activationId`).

Note this is a distinct concern from the already-flagged "generic catch clause" style issue on this same block (line 1237) — that's about catch-clause specificity; this is about what gets returned to the client.

Other new deletion endpoints added in *this same PR* (`AdminDataEndpoints`, `AdminLogsEndpoints`, `AdminFeedbackEndpoints`, `AdminMetricsEndpoints`) correctly avoid this by delegating to a `ServiceResult<T>` returned from the service layer and calling `.ToHttpResult()`, with the service catching and logging the exception and returning a generic error. Recommend converting this endpoint to the same pattern for consistency: move the try/catch into `IKnowledgeService.DeleteAllByAgentAndActivationForTenantAsync`, have it return `ServiceResult<int>`, and call `.ToHttpResult()` from the endpoint (removing the endpoint-level try/catch entirely).

XiansAi.Server.Src/Shared/Services/ActivationService.cs
hasith: **Performance:** `CleanupActivationDataAsync` runs 7 independent cleanup calls (messages, documents, knowledge, logs, schedules, app integrations, usage metrics, feedback) as fully sequential `await`s, each wrapped in its own try/catch. None of these calls depend on each other's results, so they pay the sum of 7 round-trip latencies instead of the slowest one. This method runs on *every* activation delete and is itself invoked in tight sequential loops from `AgentService` (force-delete), `TemplateService.ForceDeleteActivationsAcrossTenantsAsync`, and `TenantService.ForceDeleteActivationsForTenantAsync` — compounding the latency for bulk operations.

**Fix:** Wrap each try/catch in a local async local function/lambda and run them concurrently with `Task.WhenAll(...)`, e.g.:
```csharp
var tasks = new[]
{
    SafeCleanup(() => _messageService.DeleteMessagesByActivationAsync(tenantId, agentName, activationName), "messages"),
    SafeCleanup(() => _dataService.DeleteDocumentsByActivationAsync(tenantId, agentName, activationName), "documents"),
    // ...
};
await Task.WhenAll(tasks);
```

XiansAi.Server.Src/Shared/Services/AgentService.cs
```diff
@@ -119,20 +123,73 @@ public async Task<ServiceResult<AgentDeletionResult>> DeleteAgentAsync(string ag
...
+                    {
+                        _logger.LogWarning("Cannot delete agent {AgentName} - {Count} activation(s) exist", LogSanitizer.Sanitize(agentName), activations.Count);
+                        return ServiceResult<AgentDeletionResult>.Conflict(
+                            $"Cannot delete agent '{agentName}' because it has {activations.Count} activation(s). Please delete all activations first, or pass forceDelete=true.");
+                    }
+
+                    // Force delete: deactivate (cancels live workflows/schedules) then delete each activation first.
+                    var failedActivations = new List<string>();
+                    var deletedActivations = 0;
+                    foreach (var activation in activations)
```
hasith: **Performance:** the force-delete loop processes activations one at a time — deactivate, then delete (which itself triggers the 7-call cascade above) — so an agent with N activations makes roughly 9×N sequential round trips inside a single admin HTTP request, risking timeouts for agents with many activations.

The identical sequential pattern is duplicated in `TenantService.ForceDeleteActivationsForTenantAsync` (`TenantService.cs:1017`) and `TemplateService.ForceDeleteActivationsAcrossTenantsAsync` (`TemplateService.cs:297`) — three copies of the same loop shape.

**Fix:** Process activations with bounded concurrency (e.g. `Task.WhenAll` over batches, or a `SemaphoreSlim`-limited fan-out), aggregating results via thread-safe collections. Consider extracting the shared "deactivate + delete + cascade" step into one helper used by all three call sites, both to fix the perf issue once and to remove the triplicated logic.

XiansAi.Server.Src/Shared/Services/MessageService.cs
```diff
@@ -677,12 +678,56 @@ public async Task<ServiceResult<bool>> DeleteThreadAsync(string workflowId, stri
...
+
+    private async Task<ServiceResult<int>> DeleteMessagesAsync(Func<Task<List<ConversationThread>>> fetchCandidates)
+    {
+        try
+        {
+            var threads = await fetchCandidates();
+            var deletedCount = 0;
+            foreach (var thread in threads)
+            {
+                var deleted = await _conversationRepository.DeleteThreadAsync(thread.Id, thread.TenantId);
```
hasith: **N+1 delete pattern:** `DeleteMessagesAsync` fetches all matching conversation threads and then deletes them **one at a time** via `_conversationRepository.DeleteThreadAsync` in a loop, unlike every other repository added in this PR (`FeedbackRepository`, `KnowledgeRepository`, `LogRepository`, `UsageEventRepository` all use `DeleteManyAsync`).

**Fix:** Add a bulk `DeleteManyByFilterAsync`/`DeleteByTenantAgentAndActivationAsync` method to `ConversationRepository` using the same filter as `GetByTenantAgentAndActivationAsync`, and call it directly instead of fetch-then-loop-delete.

XiansAi.Server.Src/Shared/Repositories/FeedbackRepository.cs
```diff
@@ -189,6 +197,21 @@ public async Task<FeedbackStatsResult> GetFeedbackStatsAsync(FeedbackQuery query
...
 
+    public async Task<long> DeleteByAgentAndActivationAsync(string tenantId, string agentName, string activationName)
+    {
+        return await MongoRetryHelper.ExecuteWithRetryAsync(async () =>
+        {
+            var builder = Builders<MessageFeedbackDocument>.Filter;
+            var filter = builder.And(
+                builder.Eq(f => f.TenantId, tenantId),
+                builder.Eq(f => f.AgentName, agentName),
+                builder.Regex(f => f.WorkflowId, new BsonRegularExpression(Regex.Escape(":" + activationName) + "$")));
```
hasith: **Missing index for new delete filter:** `DeleteByAgentAndActivationAsync` filters on `tenant_id` + `agent_name` (equality) + `workflow_id` (suffix regex), but `mongodb-indexes.yaml`'s `message_feedback` section only declares `tenant_id_1_message_id_1`, `tenant_id_1_submitted_at_-1`, and `tenant_id_1_star_rating_1_submitted_at_-1` — none include `agent_name`. This delete will force a collection scan of every feedback document for the tenant.

**Fix:** Add a `tenant_id_1_agent_name_1` (or `tenant_id_1_agent_name_1_workflow_id_1`) index for `message_feedback` in `mongodb-indexes.yaml` to bound the scan before the regex is applied.

XiansAi.Server.Src/Shared/Services/ActivationService.cs
```diff
@@ -704,6 +725,10 @@ public async Task<ServiceResult<bool>> DeleteActivationAsync(string activationId
                 return ServiceResult<bool>.Conflict("Cannot delete an activation with running workflows. Please deactivate it first.");
             }
 
+            // Clean up all data scoped to this activation (app integrations, messages, documents,
+            // knowledge, logs, schedules, usage metrics, feedback) before removing the record itself.
+            await CleanupActivationDataAsync(activation);
```
hasith: `CleanupActivationDataAsync` is intentionally best-effort (documented in the method's XML comment) — the activation record is deleted regardless of whether any individual cleanup step failed, with failures only logged as warnings. This is a reasonable tradeoff to avoid blocking activation deletion, but it means cleanup failures currently have no reconciliation path: failed sub-deletes become permanently orphaned data with only a log line.

Consider surfacing per-resource cleanup failures in the delete response (e.g. `partialCleanupFailures: [...]`) or tracking them for a background reconciliation job, so orphaned data from failed cleanups isn't silently invisible.

XiansAi.Server.Src/Shared/Repositories/ConversationRepository.cs
```diff
@@ -413,6 +423,49 @@ public async Task<List<ConversationThread>> GetByTenantAndAgentAsync(string tena
...
+
+            _logger.LogDebug(
+                "GetByTenantAgentAndActivationAsync returned {Count} threads for tenant {TenantId}, agent {Agent}, activation {Activation}",
+                results.Count, LogSanitizer.Sanitize(tenantId), LogSanitizer.Sanitize(agent), LogSanitizer.Sanitize(activationName));
+
+            return results;
+        }, _logger, maxRetries: 3, baseDelayMs: 100, operationName: "GetByTenantAgentAndActivation");
+    }
+
+    public async Task<List<ConversationThread>> GetByTenantAndActivationAsync(string tenantId, string activationName)
```
hasith: `GetByTenantAndActivationAsync` (added/exposed in this PR's interface) appears to be unused dead code — no callers found anywhere in the repo. It also filters only on `tenant_id` (equality) + a suffix-regex on `workflow_id`, with no `agent` equality clause, so if it were ever called it would force a full tenant-wide collection scan (unlike `GetByTenantAgentAndActivationAsync`, which can use the `tenant_agent_lookup` index).

Recommend removing it if genuinely unused, or if kept for future use, requiring the agent name as well and documenting that it shouldn't be used as a wide, agent-less search on large tenants.

XiansAi.Server.Src/Shared/Services/AppIntegrationService.cs
```diff
@@ -709,6 +727,67 @@ public async Task<ServiceResult<bool>> DeleteBuiltinWebhookAsync(string integrat
...
+    private async Task<ServiceResult<int>> DeleteBuiltinWebhooksAsync(Func<Task<List<AppIntegration>>> fetchCandidates)
+    {
+        try
+        {
+            var toDelete = await fetchCandidates();
+            var deletedCount = 0;
+
+            foreach (var webhook in toDelete)
+            {
+                var deleteResult = await DeleteBuiltinWebhookAsync(webhook.Id, webhook.TenantId);
```
hasith: `DeleteBuiltinWebhooksAsync` (the shared private helper backing all three new bulk-webhook-delete methods) fetches candidates then deletes them one at a time sequentially, each also revoking an API key. For an agent/activation with many builtin webhooks this pays N sequential round trips.

Consider fanning out the per-webhook `DeleteBuiltinWebhookAsync` calls with bounded concurrency (e.g. `Task.WhenAll` in chunks) instead of a strictly sequential `foreach`.

XiansAi.Server.Src/Shared/Services/TenantService.cs
```diff
@@ -969,4 +986,88 @@ await _webhookEventPublisher.PublishAsync(
         }
     }
 
+    /// <summary>
+    /// Force-deletes every activation (deactivating live workflows/schedules first, then running
+    /// the full cascading data cleanup) belonging to a tenant, plus that tenant's organization-level
+    /// knowledge per agent. Best-effort — failures are logged, not thrown, so cleanup issues never
+    /// block deleting the tenant record itself.
+    /// </summary>
+    private async Task ForceDeleteActivationsForTenantAsync(string tenantId)
```
hasith: The new tenant-removal force-delete cascade (`ForceDeleteActivationsForTenantAsync`) and its template-level counterpart (`TemplateService.ForceDeleteActivationsAcrossTenantsAsync`) are irreversible, wide-blast-radius operations (all activations for a tenant, or all deployments of a template) gated by a single boolean, with no dry-run/preview mode and no logging of the invoking admin's identity (unlike `SysAdminOnlyFilter`, which does log the caller). Given the destructive scope, consider adding a `dryRun` preview that returns counts before committing, and including `tenantContext.LoggedInUser` in the summary log line for audit purposes.

PR #479, opened 2026-08-17
Improve temporal tenant isolation

Diyanada99x:


hasith:
🔍 PR Review in Progress

Claude Code is analyzing this pull request. The review will be posted here shortly.

PR Reviewer (1.9.9)

hasith (commented):
## PR Review Report

**PR:** Improve temporal tenant isolation (#479)
**Author:** hasithy
**Files Changed:** 29 | **+1138** / **-604**
**Verdict:** REQUEST CHANGES

---

### Existing review threads

- ⏳ Still open: 11 open thread(s) — no reply (avoid spam) — all from `github-code-quality` (CodeQL): missing-Dispose and generic-catch-clause findings on `TemporalGatewayService.cs`, `TenantTemporalConfigRepository.cs`, and `TenantTemporalConfigService.cs`.
- 🔇 Duplicates avoided: 1 finding (the CodeQL bot's "Missing Dispose call" on `TemporalGatewayService.cs:288` overlaps a finding independently raised by two of our reviewers — not re-posted).

---

### Summary

This PR is a substantial, well-motivated refactor: it replaces the single shared `ITemporalClientFactory`/`TemporalClientService` with a per-tenant/per-agent `ITemporalGatewayService`, adds a `TenantTemporalConfigService`/`Repository` for SysAdmin-managed per-tenant Temporal connection overrides, and introduces `Agent.OriginTenant` for tracking where a shared agent template originated. The direction is right, but the new client-caching/eviction logic has a correctness bug that undermines the credential-rotation feature it exists to support, and — more importantly — several pre-existing workflow-lookup paths still trust a caller-supplied `workflowId` without verifying tenant ownership, which is exactly the class of bug this PR is meant to close. There is also no test coverage for any of the new tenant-isolation logic or admin endpoints.

---

- [ ] `XiansAi.Server.Src/Features/WebApi/Services/WorkflowCancelService.cs:61` — `CancelWorkflow` never verifies the caller-supplied `workflowId` belongs to the caller's tenant before cancelling/terminating it, unlike the bulk-cancel path added in this same PR/file which does check `workflow.Id.StartsWith(tenantId)`. On a shared/default Temporal namespace this allows cross-tenant workflow cancellation.
  ```csharp
  // Current (problematic)
  var client = await _temporalGatewayService.GetClientAsync(_tenantContext.TenantId, WorkflowIdentifier.GetAgentName(WorkflowIdentifier.GetWorkflowType(workflowId)));
  var handle = client.GetWorkflowHandle(workflowId);

  // Fix
  if (!workflowId.StartsWith(_tenantContext.TenantId, StringComparison.Ordinal))
      return ServiceResult<WorkflowCancelResult>.Forbidden("Workflow does not belong to this tenant");
  var client = await _temporalGatewayService.GetClientAsync(_tenantContext.TenantId, WorkflowIdentifier.GetAgentName(WorkflowIdentifier.GetWorkflowType(workflowId)));
  var handle = client.GetWorkflowHandle(workflowId);
  ```

- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs:128` — `RemoveClients` only evicts cache keys prefixed `"{tenantId}:"`, missing the bare `tenantId` key created by no-agent client lookups, so rotating/reverting a tenant's Temporal credentials via the new admin endpoints doesn't take effect for those call sites; evicted clients are also never disposed (connection leak).
  ```csharp
  // Current (problematic)
  if (client.Key.StartsWith($"{tenantId}:"))
  {
      _clients.TryRemove(client.Key, out _);
  }

  // Fix
  if (client.Key == tenantId || client.Key.StartsWith($"{tenantId}:"))
  {
      if (_clients.TryRemove(client.Key, out var removed))
      {
          // dispose `removed` (async-disposable-aware) after releasing the semaphore
      }
  }
  ```

---

### Warnings (Should Fix)

- [ ] `XiansAi.Server.Src/Features/WebApi/Services/WorkflowFinderService.cs:102`, `XiansAi.Server.Src/Features/WebApi/Services/WorkflowEventsService.cs:55,206`, `XiansAi.Server.Src/Shared/Services/AdminTaskService.cs:114` *(pre-existing)* — same missing tenant-ownership check on caller-supplied `workflowId` before reading/mutating a workflow. Pre-existing, but directly relevant to this PR's isolation goal and now inconsistent with the bulk-cancel fix added here.
- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs:76` — the global `_connectionSemaphore` is acquired before checking the client cache, serializing every tenant's client fetch (even cache hits) through one lock — a contention regression vs. the old lock-free fast path.
- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs:36` and `XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs:40-42` — both constructors resolve scoped dependencies from a manually created scope that's disposed immediately, retaining instances beyond their scope's lifetime (works today only because those repositories aren't `IDisposable`).
- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs:79` — client cache keys on `tenantId:agentName` instead of the connection endpoint, so agents sharing one Temporal server each get their own connection (unbounded growth) and inflate aggregate stats in `AdminTaskService.cs:506`.
- [ ] `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminTenantEndpoints.cs:522` — `RevertTenantTemporalConfig` requires a full request body that it never reads.
- [ ] `XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs:62` — `GetForTenantAsync` returns the decrypted TLS certificate/private key verbatim in the admin GET response instead of redacting them.
- [ ] **Test coverage** — 29 changed source files, including new security-relevant components (`TenantTemporalConfigService`, `TenantTemporalConfigRepository`, the new `AdminTenantEndpoints` temporal-config routes, `TemporalGatewayService`'s tenant/agent routing, `AgentRepository.GetByNameAndOriginTenantAsync`), with zero test files added or modified.

---

- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs:182` — confirm it's intentional that resolving a client for an agent whose `OriginTenant` differs from the caller's tenant reroutes to the origin tenant's dedicated Temporal connection/credentials.
- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs:49` — unused `clients` list variable in `GetClientsAsync`.
- [ ] `XiansAi.Server.Src/Shared/Auth/TenantContext.cs:88` — `GetTemporalConfigAsync` now does an uncached DB round-trip on every call; consider a short-TTL cache.

---

### Review Details

#### Code Quality
Overall the refactor is coherent and the new `ITemporalGatewayService` interface is a reasonable consolidation of the old factory/service split. Main issues are the captive-dependency constructor pattern (duplicated across two new classes), a dead local variable, and an admin endpoint with an unused request parameter — none are hard blockers on their own, but the captive-dependency pattern is worth fixing before it becomes load-bearing.

#### Security
The standout issue is that tenant-ownership of a caller-supplied `workflowId` is still not verified before acting on it, in both a path newly touched by this PR (`WorkflowCancelService.CancelWorkflow`) and several pre-existing paths (`WorkflowFinderService`, `WorkflowEventsService`, `AdminTaskService`). This matters most on tenants without a dedicated Temporal namespace override — i.e. exactly the tenants this PR doesn't yet migrate. The decrypted cert/key echoed back from the new admin GET endpoint is a secondary hardening item. The `OriginTenant`-based rerouting deserves an explicit design confirmation since it crosses a tenant trust boundary by design (or by accident).

#### Test Coverage
No tests were added or modified for any of the new classes or endpoints in this PR. Given the change implements tenant-isolation guarantees and new SysAdmin-only endpoints handling TLS credentials, at minimum the tenant-override precedence logic in `TemporalGatewayService.GetTemporalConfig` / `TenantTemporalConfigService`, and the admin endpoints' authorization gating, should have coverage.

#### Performance
The new client cache introduces a global-lock contention point (semaphore held across cache hits, and across an `IAsyncEnumerable` consumption in `GetClientsAsync`) and drops the previous endpoint-level connection dedup, risking unbounded per-agent connection growth and skewed aggregate stats. These sit on a hot path exercised by nearly every Temporal-touching request across all tenants and should be addressed before this scales up.

---

### Files Reviewed

| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs` | +332/-0 (new) | 🔴 High | New hot-path client cache; locking, eviction, dedup issues |
| `XiansAi.Server.Src/Features/WebApi/Services/WorkflowCancelService.cs` | ~+61/-40 | 🔴 High | Missing tenant-ownership check on single-cancel path |
| `XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs` | +220/-0 (new) | 🔴 High | New SysAdmin credential management service |
| `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminTenantEndpoints.cs` | +84/-0 | 🟡 Medium | New SysAdmin endpoints for per-tenant Temporal override |
| `XiansAi.Server.Src/Shared/Repositories/TenantTemporalConfigRepository.cs` | +143/-0 (new) | 🟡 Medium | Encrypted credential storage |
| `XiansAi.Server.Src/Shared/Auth/TenantContext.cs` | ~+34/-? | 🟡 Medium | Async tenant config resolution, uncached |
| `XiansAi.Server.Src/Shared/Services/AdminTaskService.cs` | ~+117/-? | 🟡 Medium | Multi-client aggregation, ownership check gap |
| `XiansAi.Server.Src/Features/WebApi/Services/WorkflowFinderService.cs` | ~+51/-? | 🟡 Medium | Ownership check gap |
| `XiansAi.Server.Src/Features/WebApi/Services/WorkflowEventsService.cs` | ~+10/-? | 🟡 Medium | Ownership check gap |
| `XiansAi.Server.Src/Shared/Data/Models/Agent.cs` | +9/-? | 🟡 Medium | New `OriginTenant` property |
| `XiansAi.Server.Src/Shared/Repositories/AgentRepository.cs` | +7/-? | 🟢 Low | New `GetByNameAndOriginTenantAsync` |
| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientService.cs` | -352 (removed) | 🟢 Low | Superseded by TemporalGatewayService |
| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientFactory.cs` | -34 (removed) | 🟢 Low | Superseded |
| `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayFactory.cs` | +39 (new) | 🟢 Low | Thin factory wrapper |
| …and 15 more files | +~93/-~118 | 🟢 Low | Call-site updates to the new gateway API (`ScheduleService`, `TaskService`, `WebhookService`, `ActivationCleanupService`, `CertificateService`, `WorkflowSignalService`, `UpdateService`, `Program.cs`, DI registration, config) |

XiansAi.Server.Src/Features/WebApi/Services/WorkflowCancelService.cs
```diff
@@ -58,7 +58,7 @@ public async Task<ServiceResult<WorkflowCancelResult>> CancelWorkflow(string wor
                 return ServiceResult<WorkflowCancelResult>.BadRequest("WorkflowId is required");
             }
 
-            var client = await _temporalClientFactory.GetClientAsync();
+            var client = await _temporalGatewayService.GetClientAsync(_tenantContext.TenantId, WorkflowIdentifier.GetAgentName(WorkflowIdentifier.GetWorkflowType(workflowId)));
```
hasith: **CRITICAL — missing tenant-ownership check on `workflowId` (cross-tenant cancel/terminate)**

`CancelWorkflow` picks a Temporal client scoped to the caller's own tenant/agent, but never verifies the caller-supplied `workflowId` actually belongs to that tenant before calling `client.GetWorkflowHandle(workflowId)` and cancelling/terminating it. Notice the bulk path in this same file (`CancelAllWorkflowsForTenant`, added by this PR) does check `workflow.Id.StartsWith(tenantId)` before acting — the single-workflow path lacks the equivalent guard. On any tenant still using the shared/default Temporal namespace (i.e. no dedicated per-tenant override configured via the new admin endpoints), an authenticated user from Tenant A can supply Tenant B's workflow ID and cancel/terminate Tenant B's running workflow.

**Fix:** validate that `workflowId` is prefixed with (or otherwise resolves to) `_tenantContext.TenantId` before calling `GetWorkflowHandle`, mirroring the check already added in `CancelAllWorkflowsForTenant`.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs
```diff
@@ -0,0 +1,332 @@
...
+
+    public async Task RemoveClients(string tenantId)
+    {
+        ThrowIfDisposed();
+        await _connectionSemaphore.WaitAsync();
+        try
+        {
+            foreach (var client in _clients)
+            {
+                if (client.Key.StartsWith($"{tenantId}:"))
```
hasith: **CRITICAL — `RemoveClients` doesn't evict the no-agent cache entry, and never disposes evicted clients**

`RemoveClients(tenantId)` only removes cache keys matching `"{tenantId}:"` (agent-scoped keys). But `GetClientInternalAsync` also caches a bare-`tenantId` entry whenever `agentName` is empty (`clientKey = string.IsNullOrEmpty(agentName) ? tenantId : $"{tenantId}:{agentName}"`, line 79) — that key never starts with `"{tenantId}:"` so it's never removed. Since `RemoveClients` is called from `TenantTemporalConfigService.UpsertAsync`/`RevertAsync` whenever a SysAdmin rotates or reverts a tenant's Temporal credentials, callers that use the no-agent `GetClientAsync`/`GetClientInternalAsync` overload keep using the stale, previously-cached connection/credentials indefinitely — defeating the purpose of the new revert/rotate feature. Additionally, the clients that *are* removed are only `TryRemove`'d, never `Dispose`d/`DisposeAsync`'d, leaking the underlying Temporal/gRPC connection on every tenant reconfiguration.

**Fix:** match on `client.Key == tenantId || client.Key.StartsWith($"{tenantId}:")`, and dispose each removed client (async-disposable-aware, as already done in `DisposeAsyncCore`).

XiansAi.Server.Src/Features/WebApi/Services/WorkflowFinderService.cs
```diff
@@ -99,7 +99,7 @@ public async Task<ServiceResult<WorkflowResponse>> GetWorkflow(string workflowId
         try
         {
             _logger.LogInformation("Retrieving workflow with ID: {WorkflowId} and workflowRunId: {WorkflowRunId}", LogSanitizer.Sanitize(workflowId), LogSanitizer.Sanitize(workflowRunId));
-            var client = await _clientFactory.GetClientAsync();
+            var client = await _temporalGatewayFactory.GetClientAsync(WorkflowIdentifier.GetAgentName(WorkflowIdentifier.GetWorkflowType(workflowId)));
```
hasith: **WARNING (pre-existing, but relevant to this PR's stated goal) — no tenant-ownership check on caller-supplied `workflowId`**

`GetWorkflow` resolves a client from the agent name embedded in `workflowId` and calls `client.GetWorkflowHandle(workflowId, ...)` directly, with no check that the workflow actually belongs to `_tenantContext.TenantId`. The same gap exists in `WorkflowEventsService.cs` (`StreamWorkflowEvents`/`GetWorkflowEvents`) and `AdminTaskService.cs` (`GetTaskById` and the task-mutation methods). This predates this PR, but this PR explicitly adds a tenant-ownership check to the bulk-cancel path in `WorkflowCancelService.cs` — leaving these read/mutate paths without an equivalent guard is an inconsistent, incomplete fix for the isolation problem this PR is meant to solve, and on a shared/default Temporal namespace it allows cross-tenant workflow read/update access via a guessed or leaked workflow ID.

**Fix:** add the same `workflowId` tenant-ownership check used in the new bulk-cancel path to these read/mutate call sites.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    public async Task<ITemporalClient> GetClientInternalAsync(string tenantId, string? agentName)
+    {
+        ThrowIfDisposed();
+
+        if (string.IsNullOrEmpty(tenantId))
+            throw new InvalidOperationException("TenantId is required");
+
+        await _connectionSemaphore.WaitAsync();
```
hasith: **WARNING — global semaphore now serializes every client fetch, including cache hits**

`GetClientInternalAsync` acquires the single app-wide `_connectionSemaphore` (line 76) *before* checking the `_clients` cache (line 80), so every tenant's client lookup — even a pure cache hit — is now serialized behind one lock for the whole process. The previous `TemporalClientService` had a lock-free `TryGetValue` fast path before taking any lock. Under load, a slow cache-miss connect (network `TemporalClient.ConnectAsync` + DB config lookup) for one tenant now blocks client lookups for every other tenant too.

**Fix:** check the cache before acquiring the semaphore (double-checked locking), only taking the lock on a miss.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs
```diff
@@ -0,0 +1,332 @@
...
+    private readonly IConfiguration _configuration;
+    private readonly ILogger<TemporalGatewayService> _logger;
+
+    public TemporalGatewayService(
+        IServiceScopeFactory serviceFactory,
+        ILogger<TemporalGatewayService> logger,
+        IConfiguration configuration)
+    {
+        ArgumentNullException.ThrowIfNull(serviceFactory);
+        using var scope = serviceFactory.CreateScope();
```
hasith: **WARNING — captive-dependency pattern in constructor**

`TemporalGatewayService` is registered `Singleton` (`SharedServices.cs`), but its constructor creates a temporary DI scope, resolves the `Scoped` `ITenantTemporalConfigRepository`/`IAgentRepository` from it, and disposes the scope immediately afterward — retaining references to instances whose owning scope no longer exists. The same pattern also appears in `TenantTemporalConfigService`'s constructor (`XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs:40-42`). It happens to work today only because neither `AgentRepository` nor `TenantTemporalConfigRepository` implements `IDisposable`/`IAsyncDisposable` (so the DI container doesn't dispose them when the temporary scope is disposed) — but it's fragile: if either repository implementation ever becomes disposable, every call through the retained reference throws `ObjectDisposedException`.

**Fix:** for the singleton, keep `IServiceScopeFactory` and create/dispose a scope per operation instead of retaining scoped instances; for the scoped `TenantTemporalConfigService`, just inject `ITenantTemporalConfigRepository`/`ITemporalGatewayService` directly via the constructor rather than resolving them from a manually created scope.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs
```diff
@@ -0,0 +1,332 @@
...
+    {
+        ThrowIfDisposed();
+
+        if (string.IsNullOrEmpty(tenantId))
+            throw new InvalidOperationException("TenantId is required");
+
+        await _connectionSemaphore.WaitAsync();
+        try
+        {
+            String clientKey = string.IsNullOrEmpty(agentName) ? tenantId : $"{tenantId}:{agentName}";
```
hasith: **WARNING — per-agent client caching removes endpoint-level connection dedup**

The previous `TemporalClientService` deduplicated connections across an endpoint/namespace/TLS fingerprint; this cache now keys strictly on `tenantId:agentName` (line 79). Since most agents within a tenant share the same default Temporal server/namespace, every distinct agent now opens and holds its own separate `TemporalClient`/gRPC connection instead of reusing one — unbounded connection growth as more agents are used per tenant. It also skews aggregation: `AdminTaskService.cs:506` sums stats across every client returned by `GetClientsAsync(tenantId)`, so a tenant with N agents on the same shared server now gets counts/queries inflated up to Nx.

**Fix:** key the cache (or a secondary lookup) by the resolved connection target (server URL + namespace + TLS fingerprint) so agents sharing an endpoint reuse one client.

XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminTenantEndpoints.cs
```diff
@@ -452,6 +452,90 @@ public static void MapAdminTenantEndpoints(this RouteGroupBuilder adminApiGroup)
...
+    {
+        var actor = tenantContext.LoggedInUser ?? "system";
+        var result = await service.UpsertAsync(
+            tenantId, request.ServerUrl, request.Namespace, request.Certificate, request.PrivateKey, actor);
+        return result.ToHttpResult();
+    }
+
+    private static async Task<IResult> RevertTenantTemporalConfig(
+        string tenantId,
+        [FromBody] UpsertTenantTemporalConfigRequest request,
```
hasith: **WARNING — unused request body parameter on revert endpoint**

`RevertTenantTemporalConfig` requires callers to submit a full `UpsertTenantTemporalConfigRequest` body (`[FromBody] UpsertTenantTemporalConfigRequest request`), including its `required` fields (`TenantId`, `ServerUrl`, `Namespace`), but never reads `request` — the method only uses the `tenantId` route parameter. This forces SysAdmin API callers to fabricate a meaningless payload just to invoke a revert, and is confusing/error-prone API design.

**Fix:** drop the `request` parameter (and its `[FromBody]` binding) from the revert endpoint signature — a POST with no body is sufficient for a revert action.

XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs
```diff
@@ -0,0 +1,220 @@
...
+        {
+            // Repository only returns the active (non-deleted) row, already decrypted.
+            var doc = await _repository.GetAsync(tenantId);
+            if (doc == null) return ServiceResult<UpsertTenantTemporalConfigRequest?>.Success(null);
+
+            var config = new UpsertTenantTemporalConfigRequest
+            {
+                TenantId = doc.TenantId,
+                ServerUrl = doc.ServerUrl,
+                Namespace = doc.Namespace,
```
hasith: **WARNING — decrypted TLS client certificate/private key returned verbatim in admin API response**

`GetForTenantAsync` maps the repository's already-decrypted `Certificate`/`PrivateKey` straight into the response DTO (lines 62-63), which is returned as-is by `AdminGetTenantTemporalConfig` (`GET /tenants/{tenantId}/temporal-config`). Even though this endpoint is SysAdmin-gated, returning raw private key material in a JSON response body unnecessarily widens the exposure surface (browser history, proxies/gateways, request logging, audit trails) compared to a redacted/masked placeholder (e.g. `HasCredentials: true` + a fingerprint).

**Fix:** omit `PrivateKey` (and ideally `Certificate`) from the GET response, or replace with a redacted indicator, and only accept them on write (PUT/POST).

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs
```diff
@@ -0,0 +1,332 @@
...
+                "If workflows fail to start, manually register these attributes: {Attributes}",
+                LogSanitizer.Sanitize(@namespace), string.Join(", ", Constants.RequiredSearchAttributes.Keys));
+        }
+    }
+    private async Task<TemporalConfig> GetTemporalConfig(string tenantId, string? agentName)
+    {
+
+        if (!string.IsNullOrEmpty(agentName))
+        {
+            var agent = await _agentRepository.GetByNameAndOriginTenantAsync(agentName, tenantId);
```
hasith: **SUGGESTION — confirm intentional: agent client resolution silently reroutes to the agent's origin tenant**

When resolving a Temporal client for `(tenantId, agentName)`, if the agent's `OriginTenant` differs from the caller's `tenantId`, the code reassigns `tenantId = agent.OriginTenant` (line 182-186) before looking up the Temporal connection — so the connection/credentials used are the *origin* tenant's, not the deploying tenant's. If a shared/system agent template can be deployed across tenants, this routes that agent's workflow traffic through, and makes it visible within, the origin tenant's private Temporal namespace and credentials. Please confirm this cross-tenant routing is deliberate (e.g. for shared system agents) rather than an oversight, since it's a meaningful trust-boundary decision.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs
```diff
@@ -0,0 +1,332 @@
...
+        _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+    }
+
+    public async IAsyncEnumerable<ITemporalClient> GetClientsAsync(string tenantId)
+    {
+        ThrowIfDisposed();
+        await _connectionSemaphore.WaitAsync();
+        try
+        {
+            var clients = new List<ITemporalClient>();
```
hasith: **SUGGESTION — unused local variable**

`var clients = new List<ITemporalClient>();` is declared in `GetClientsAsync` but never populated or read — the method `yield return`s directly from the iteration instead. Dead code left over from a refactor.

**Fix:** remove the unused `clients` declaration.

```suggestion
            foreach (var client in _clients)
```

XiansAi.Server.Src/Shared/Auth/TenantContext.cs
```diff
@@ -71,17 +73,31 @@ public string ParticipantId
...
-        { 
-            if (string.IsNullOrEmpty(TenantId)) 
+        public async Task<TemporalConfig> GetTemporalConfigAsync()
+        {
+            if (string.IsNullOrEmpty(TenantId))
                  throw new InvalidOperationException("TenantId is required");
 
-            // get the temporal config for the tenant
+            // A tenant-specific override saved via the admin UI takes precedence.
+            var overrideConfig = await _tenantTemporalConfigRepository.GetAsync(TenantId);
```
hasith: **SUGGESTION — uncached per-call DB lookup for tenant Temporal override**

`GetTemporalConfigAsync` now performs a MongoDB round-trip (`_tenantTemporalConfigRepository.GetAsync(TenantId)`) on every call, even though the large majority of tenants have no override configured. Call sites like `CertificateService.GetFlowServerSettingsAsync` and workflow polling paths pay this extra latency on every invocation, with no caching layer (unlike the connection-level cache in `TemporalGatewayService`).

**Fix:** consider a short-TTL cache (or a negative-result cache) for the per-tenant override lookup, invalidated by `TenantTemporalConfigService.UpsertAsync`/`RevertAsync`.

hasith:
🔍 PR Review in Progress

Claude Code is analyzing this pull request. The review will be posted here shortly.

PR Reviewer (1.9.9)

hasith (commented):
## PR Review Report

**PR:** Improve temporal tenant isolation
**Author:** Hasith Yaggahavita
**Files Changed:** 36 | **+1958** / **-724**
**Verdict:** REQUEST CHANGES

---

### Re-review delta
Reviewed 2 new commit(s) since the last review (`723f931b`..`af280b1`).
- ✅ Fixed: 4 previously-flagged issue(s) resolved
- ⏳ Still open: 7 carried-over issue(s)
- 🆕 New: 5 issue(s) introduced since the last review

Notably, the earlier CRITICAL finding about `TemporalGatewayService.RemoveClients` never disposing evicted Temporal clients, the captive-dependency constructor pattern, the per-agent connection-dedup gap, and the unused `clients` variable have all been verified as fixed in this revision (`RemoveClients` now calls `DisposeClientsAsync`; `GetClientsAsync` dedupes by connection identity; DI scopes are created/disposed per call).

---

### Existing review threads
- ✅ Appears addressed: 1 open thread (CodeQL "Missing Dispose call on local IDisposable" for a `CancellationTokenSource` in the now-deleted `TemporalClientService.cs`) — will reply, leave open for original author
- ⏳ Still open: 10 open thread(s) (CodeQL generic-catch-clause / log-injection / container-not-accessed findings) — no reply (avoid spam)
- 🔇 Duplicates avoided: 0 findings not re-posted

---

### Summary
This PR replaces the single-client `TemporalClientFactory`/`TemporalClientService` with a per-tenant/per-agent `TemporalGatewayFactory`/`TemporalGatewayService` + `TemporalClientCache`, adds `Agent.OriginTenant`, a `TenantTemporalConfigService` for per-tenant Temporal server overrides, and new SysAdmin endpoints to manage them. Since the last review, the connection-cache correctness issues (client disposal on eviction, captive-dependency DI, connection dedup) have been substantially fixed. However, a genuine cross-tenant IDOR remains open in `WorkflowCancelService.CancelWorkflow` (carried over from the last review), and this round surfaces a **new, equally serious IDOR** in `ScheduleService`'s three read paths (`GetUpcomingRunsAsync`, `GetScheduleHistoryAsync`, `GetScheduleByIdAsync`) — they resolve schedules across every origin-tenant Temporal cluster but never check the resolved schedule's `TenantId` against the caller's tenant, unlike the write paths in the same file which already do this correctly.

---

- [ ] `XiansAi.Server.Src/Features/WebApi/Services/ScheduleService.cs:232` (also :277, :325) — **NEW.** `GetUpcomingRunsAsync`, `GetScheduleHistoryAsync`, and `GetScheduleByIdAsync` resolve a client via `FindClientForScheduleAsync` (which now searches every origin-tenant Temporal cluster) and describe the schedule with no check that its `TenantId` matches the caller's tenant — a cross-tenant read (IDOR). The write paths in the same file (`DeleteScheduleAsync`, `PauseScheduleAsync`, `ResumeScheduleAsync`) already guard this via `AuthorizeScheduleWriteAsync`.
  ```csharp
  // Current (problematic) — GetScheduleByIdAsync
  var client = await FindClientForScheduleAsync(scheduleId);
  var scheduleHandle = client.GetScheduleHandle(scheduleId);
  var description = await scheduleHandle.DescribeAsync();
  var schedule = MapToScheduleModel(scheduleId, description);
  return ServiceResult<ScheduleModel>.Success(schedule);

  // Fix — reject cross-tenant schedules, mirroring AuthorizeScheduleWriteAsync
  var schedule = MapToScheduleModel(scheduleId, description);
  if (!string.IsNullOrEmpty(schedule.TenantId) && schedule.TenantId != _tenantContext.TenantId)
  {
      return ServiceResult<ScheduleModel>.NotFound($"Schedule {scheduleId} not found");
  }
  return ServiceResult<ScheduleModel>.Success(schedule);
  ```

- [ ] `XiansAi.Server.Src/Features/WebApi/Services/WorkflowCancelService.cs:70` — **Carried over from the last review, still unresolved by the plugin's own reconciliation check.** *(Lead's note: manual inspection shows a `WorkflowBelongsToCurrentTenant(workflowId)` guard has since been added at line 61 of the current file, which appears to functionally address this. It remains flagged here because the mechanical fid-matching used for re-review intentionally errs conservative — the originally-flagged `GetClientAsync(...)` statement is still present, just now preceded by the new guard — so please confirm the fix is complete and reply/resolve on the original thread once verified.)*

*(2 critical issues found — must be resolved before merge.)*

---

### Warnings (Should Fix)

- [ ] `XiansAi.Server.Src/Features/WebApi/Services/ScheduleService.cs:361` — **NEW.** `FindClientForScheduleAsync` calls `DescribeAsync()` to probe cluster ownership, then every caller calls `DescribeAsync()` again — doubling Temporal RPC round-trips on every single-schedule get/pause/resume/delete. Return the already-fetched description from the probe and reuse it.
- [ ] `XiansAi.Server.Src/Shared/Services/AdminTaskService.cs:557` — **NEW.** `GetTaskStatistics` resolves the tenant's client list twice (once in `GetTaskStatisticsUsingCountApi`, again in `GetTimedOutCountUsingList`), each re-triggering an origin-tenant DB lookup and cluster fan-out. Resolve once and pass the list to both.
- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs:278` — **NEW.** `ResolveTemporalConfigTenantIdsAsync` issues an uncached `AgentRepository.GetDistinctOriginTenantsAsync` Mongo query on every `GetClientsAsync` call (a hot path used by Schedule/Workflow/Task services). Cache the per-tenant origin-tenant list with invalidation on agent/config changes.
- [ ] `XiansAi.Server.Src/Features/WebApi/Services/WorkflowFinderService.cs:102` — carried over (pre-existing, same class of tenant-ownership gap as the `WorkflowCancelService`/`ScheduleService` issues above — worth fixing as part of this PR's stated goal).
- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs:64` — carried over — global semaphore serializes all client fetches including cache hits; check the cache before acquiring the lock.
- [ ] `XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminTenantEndpoints.cs:522` — carried over — `RevertTenantTemporalConfig` requires a full request body it never reads.
- [ ] `XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs:61` — carried over — `GetForTenantAsync` returns decrypted certificate/private key material verbatim to the admin GET endpoint.

---

- [ ] `XiansAi.Server.Tests/UnitTests/Shared/Utils/TemporalGatewayServiceTests.cs:100` — **NEW.** No regression test pins down that `RemoveClients` disposes evicted clients at the service layer (only cache-level eviction is tested) — add one so a future refactor can't silently reintroduce the leak that was just fixed.
- [ ] `XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs` (origin-tenant reroute) — carried over — confirm the silent reroute of agent client resolution to `agent.OriginTenant` is intentional, and document the trust boundary.
- [ ] `XiansAi.Server.Src/Shared/Auth/TenantContext.cs:88` — carried over — `GetTemporalConfigAsync` performs an uncached DB round-trip per call.

---

### Review Details

#### Code Quality
The refactor from `ITemporalClientFactory`/`TemporalClientService` to `ITemporalGatewayFactory`/`TemporalGatewayService` + `TemporalClientCache` is a clean separation of concerns, and prior structural issues (captive-dependency DI, connection dedup, unused variables) have been fixed since the last review. Remaining nits are the unread request body on the admin revert endpoint (carried over).

#### Security
The dominant concern is tenant-isolation itself, which is this PR's whole purpose. `WorkflowCancelService.CancelWorkflow` (carried over) and the newly-discovered `ScheduleService` read paths both allow schedule/workflow data belonging to another tenant to be accessed given a guessable/leaked id, and `FindClientForScheduleAsync`'s origin-tenant cluster search widens the blast radius of that gap. `WorkflowFinderService.cs:102` has the same pre-existing gap. Decrypted TLS credentials are also returned verbatim from the admin GET config endpoint (carried over).

#### Test Coverage
New unit tests cover `TemporalClientCache` eviction/dedup, `TemporalGatewayFactory`, `TemporalGatewayService`, and `WorkflowCancelService`'s tenant-check addition — good coverage of the areas that changed. Gaps: no test locks in the client-disposal-on-eviction fix at the service layer, and there's no regression test for either the carried-over `WorkflowCancelService` IDOR or the new `ScheduleService` IDOR.

#### Performance
The connection-cache locking/dedup issues raised previously are resolved. New findings this round are all duplicate-RPC / duplicate-DB-lookup regressions introduced by the multi-cluster search this PR adds (`FindClientForScheduleAsync`, `AdminTaskService.GetTaskStatistics`, `ResolveTemporalConfigTenantIdsAsync`) — none are correctness-breaking, but all scale with the number of origin-tenant clusters a tenant has.

---

### Files Reviewed

| XiansAi.Server.Src/Features/WebApi/Services/ScheduleService.cs | +139/-76 | 🔴 High | New IDOR + duplicate-RPC findings |
| XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs | new (422) | 🔴 High | Core of the refactor; most prior findings now fixed |
| XiansAi.Server.Src/Features/WebApi/Services/WorkflowCancelService.cs | +51/-26 | 🔴 High | Carried-over IDOR on single-workflow cancel |
| XiansAi.Server.Src/Shared/Services/AdminTaskService.cs | +117/-84 | 🟡 Medium | Redundant client-list resolution |
| XiansAi.Server.Src/Shared/Services/TenantTemporalConfigService.cs | new (219) | 🟡 Medium | Secrets exposure on GET (carried over) |
| XiansAi.Server.Src/Features/AdminApi/Endpoints/AdminTenantEndpoints.cs | +84 | 🟡 Medium | Unused request body (carried over) |
| XiansAi.Server.Src/Shared/Utils/Temporal/TemporalClientCache.cs | new (105) | 🟢 Low | Eviction/dedup fixed since last review |
| XiansAi.Server.Src/Shared/Auth/TenantContext.cs | +38/-8 | 🟢 Low | Uncached lookup (carried over, suggestion) |
| XiansAi.Server.Src/Features/WebApi/Services/WorkflowFinderService.cs | +81/-... | 🟡 Medium | Carried-over tenant-ownership gap |
| …and 27 more files (tests, remaining services, config) | +~1400/-~500 | 🟢 Low | Mostly test coverage and mechanical call-site updates |

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs
hasith: ✅ Resolved as of `af280b16bff61ee218c4149aff4797bdf9364ca1`. This finding no longer reproduces against the current head.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs
hasith: ✅ Resolved as of `af280b16bff61ee218c4149aff4797bdf9364ca1`. This finding no longer reproduces against the current head.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs
hasith: ✅ Resolved as of `af280b16bff61ee218c4149aff4797bdf9364ca1`. This finding no longer reproduces against the current head.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs
hasith: ✅ Resolved as of `af280b16bff61ee218c4149aff4797bdf9364ca1`. This finding no longer reproduces against the current head.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs
```diff
@@ -0,0 +1,329 @@
...
+        lock (_disposeLock)
+        {
+            if (_disposed) return;
+            _disposed = true;
+        }
+
+        _logger.LogInformation("Disposing Temporal client service asynchronously");
+
+        var disposeTimeout = TimeSpan.FromSeconds(10);
+        var cancellationTokenSource = new CancellationTokenSource(disposeTimeout);
```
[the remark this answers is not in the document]
hasith: Looks addressed as of `af280b16bff61ee218c4149aff4797bdf9364ca1` — leaving this thread open for the original author to resolve.

XiansAi.Server.Src/Features/WebApi/Services/ScheduleService.cs
```diff
@@ -221,7 +229,11 @@ public async Task<ServiceResult<List<ScheduleRunModel>>> GetUpcomingRunsAsync(
     {
         try
         {
-            var client = await _clientFactory.GetClientAsync();
+            var client = await FindClientForScheduleAsync(scheduleId);
```
hasith: **CRITICAL — missing tenant-ownership check on schedule read paths (cross-tenant IDOR)**

`GetUpcomingRunsAsync`, `GetScheduleHistoryAsync` (line 277), and `GetScheduleByIdAsync` (line 325) all resolve a Temporal client via the new `FindClientForScheduleAsync(scheduleId)` helper and then call `DescribeAsync()` directly on the caller-supplied `scheduleId` — with no check that the schedule's `TenantId` (extracted from its memo, see `MapToScheduleModel`) matches `_tenantContext.TenantId`.

Contrast this with the mutation paths in the same file (`DeleteScheduleAsync`, `PauseScheduleAsync`, `ResumeScheduleAsync`) which correctly call `AuthorizeScheduleWriteAsync` right after obtaining the handle, rejecting cross-tenant schedules (see `scheduleModel.TenantId != _tenantContext.TenantId` around line 1063). The three read-only methods above lack an equivalent guard, so any authenticated user who knows or can guess another tenant's `scheduleId` can read its upcoming runs, run history, and full schedule detail (agent name, workflow type, cron spec).

This is made worse by this PR's own change: `FindClientForScheduleAsync` now searches *every* Temporal cluster reachable via the tenant's agents' `OriginTenant` (not just the tenant's own default client), so this IDOR can reach schedules on dedicated per-tenant Temporal clusters that were previously unreachable from another tenant's context.

**Fix:** After `DescribeAsync()` in all three read methods, map to `ScheduleModel` and reject (404/403) when `scheduleModel.TenantId` is set and differs from `_tenantContext.TenantId` — the same check `AuthorizeScheduleWriteAsync` already implements for the write paths. Given this PR's stated goal is tenant isolation, this should be fixed before merge.

XiansAi.Server.Src/Features/WebApi/Services/ScheduleService.cs
```diff
@@ -328,6 +348,34 @@ public async Task<ServiceResult<ScheduleModel>> GetScheduleByIdAsync(string sche
...
+    /// Locates the Temporal client that hosts <paramref name="scheduleId"/>, searching every
+    /// cluster the current tenant may use (the tenant's own config plus origin-tenant clusters).
+    /// </summary>
+    private async Task<ITemporalClient?> FindClientForScheduleAsync(string scheduleId)
+    {
+        await foreach (var client in _temporalGatewayFactory.GetClientsAsync(_tenantContext.TenantId))
+        {
+            try
+            {
+                await client.GetScheduleHandle(scheduleId).DescribeAsync();
```
hasith: **WARNING — duplicate `DescribeAsync` RPC on every single-schedule operation**

`FindClientForScheduleAsync` calls `client.GetScheduleHandle(scheduleId).DescribeAsync()` purely to probe which Temporal cluster owns the schedule (line 361), and then every caller (`GetUpcomingRunsAsync` line 238, `GetScheduleHistoryAsync` line 283, `GetScheduleByIdAsync` line 331, plus the write paths via `AuthorizeScheduleWriteAsync`) immediately calls `DescribeAsync()` again on the same schedule right after.

Every single-schedule read or mutation now issues at least 2 sequential Temporal `DescribeSchedule` RPCs instead of 1, and when a tenant has multiple origin-tenant clusters, the probe loop in `FindClientForScheduleAsync` performs up to N sequential RPCs before reaching the "real" describe call — directly doubling (or worse) round-trip latency on what used to be single-call operations.

**Fix:** Have `FindClientForScheduleAsync` return the already-fetched `ScheduleDescription` (or a `(client, description)` tuple) alongside the client, and have callers reuse that description instead of calling `DescribeAsync()` a second time.

XiansAi.Server.Src/Shared/Services/AdminTaskService.cs
```diff
@@ -535,7 +554,7 @@ public async Task<ServiceResult<TaskStatisticsResponse>> GetTaskStatistics(
             }
 
             // For timed out tasks, we still need to query individual workflows since it's stored in memo
-            var timedOut = await GetTimedOutCountUsingList(client, baseQuery);
+            var timedOut = await GetTimedOutCountUsingList(tenantId, baseQuery);
```
hasith: **WARNING — redundant client-list resolution in `GetTaskStatistics`**

`GetTaskStatisticsUsingCountApi` resolves the tenant's Temporal client list at line 518 (`await foreach (var client in _temporalGatewayFactory.GetClientsAsync(tenantId))`), then calls `GetTimedOutCountUsingList` at line 557, which independently re-resolves the *same* client list again at line 597. Each resolution triggers a fresh `AgentRepository.GetDistinctOriginTenantsAsync` Mongo query (see `TemporalGatewayService.ResolveTemporalConfigTenantIdsAsync`) plus per-cluster client lookups.

Every call to the task-statistics endpoint does roughly 2x the DB round-trips and cluster fan-outs necessary. Contrast with `GetTasks` in the same file, which resolves the client list once and passes it into `GetWorkflowCountAsync`.

**Fix:** Resolve `_temporalGatewayFactory.GetClientsAsync(tenantId)` into a single list once at the top of `GetTaskStatistics`, and pass it into both `GetTaskStatisticsUsingCountApi` and `GetTimedOutCountUsingList` instead of having each re-resolve it.

XiansAi.Server.Src/Shared/Utils/Temporal/TemporalGatewayService.cs
```diff
@@ -0,0 +1,422 @@
...
+    private async Task<List<string>> ResolveTemporalConfigTenantIdsAsync(string tenantId)
+    {
+        var configTenantIds = new List<string> { tenantId };
+        var seen = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { tenantId };
+
+        try
+        {
+            using var scope = _serviceScopeFactory.CreateScope();
+            var agentRepository = scope.ServiceProvider.GetRequiredService<IAgentRepository>();
+            var originTenants = await agentRepository.GetDistinctOriginTenantsAsync(tenantId) ?? new List<string>();
```
hasith: **WARNING — uncached origin-tenant lookup on every `GetClientsAsync` call**

`ResolveTemporalConfigTenantIdsAsync` (called on every `GetClientsAsync` invocation — used by `ScheduleService`, `WorkflowFinderService`, `WorkflowCancelService.CancelAllWorkflowsForTenant`, and `AdminTaskService` statistics) issues an uncached `agentRepository.GetDistinctOriginTenantsAsync(tenantId)` Mongo query every single time, even though which tenants an agent originates from changes rarely.

This is a distinct hot-path DB round-trip from the connection-level cache already in `TemporalClientCache` — it sits in front of the cache lookup, so even a fully cached connection pays this query cost on every request.

**Fix:** Cache the resolved origin-tenant id list per tenant (e.g. a short-TTL `IMemoryCache`, or stored alongside `TemporalClientCache`), invalidated when agents are created/updated or when `RemoveClients` is invoked for that tenant.

XiansAi.Server.Tests/UnitTests/Shared/Utils/TemporalGatewayServiceTests.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        var service = CreateService(agentRepository.Object, temporalConfigRepository.Object);
+
+        var enumerator = service.GetClientsAsync(TenantId).GetAsyncEnumerator();
+        await Assert.ThrowsAsync<InvalidOperationException>(
+            async () => await enumerator.MoveNextAsync());
+
+        temporalConfigRepository.Verify(r => r.GetAsync(TenantId), Times.Once);
+        temporalConfigRepository.Verify(r => r.GetAsync("platform"), Times.Once);
+    }
```
hasith: **SUGGESTION — no regression test for `RemoveClients` disposing evicted Temporal clients**

`TemporalClientCacheTests.cs` already covers cache-level eviction (`RemoveByTenant_EvictsBareTenantDefaultClient`, `RemoveByTenant_EvictsOriginRoutedKeys_FromOtherTenants`), but `TemporalGatewayServiceTests.cs` has no test exercising `TemporalGatewayService.RemoveClients(tenantId)` itself — specifically that it calls `DisposeClientsAsync` on the clients returned by `_clientCache.RemoveByTenant`. This is the exact bug an earlier review round on this PR flagged (evicted clients never disposed); it's since been fixed in `RemoveClients` (line 136: `await DisposeClientsAsync(clientsToDispose, ...)`), but nothing pins that behavior down at the service layer, so a future refactor could silently reintroduce the leak.

**Fix:** Add a test that injects a fake/mock `ITemporalClient` into the cache, calls `RemoveClients(tenantId)`, and asserts the client's dispose method was invoked.

